PR #40, opened 2024-05-10
[DO NOT MERGE] Update app_signals to application_signals

bjrara:
*Issue #, if available:*

*Description of changes:*
Rename app_signals to application_signals for GA release. It's part of the PR: https://github.com/aws/amazon-cloudwatch-agent/pull/1133.


charts/amazon-cloudwatch-observability/templates/_helpers.tpl
```diff
@@ -17,8 +17,8 @@ Helper function to modify cloudwatch-agent config
 {{- $agent := set $configCopy "agent" $agentRegion }}
 {{- end }}
 
-{{- $appSignals := pluck "app_signals" $configCopy.logs.metrics_collected | first }}
-{{- if and (hasKey $configCopy.logs.metrics_collected "app_signals") (empty $appSignals.hosted_in) }}
+{{- $appSignals := pluck "application_signals" $configCopy.logs.metrics_collected | first }}
```
mitali-salvi, line 20: Should we be supporting both names ? are we deprecating app_signals altogether ?

bjrara: Since helm is a GA feature, I don't think we need to be compatible with `app_signals`. It is deprecated. We will always recommend using `application_signals` in any case.

mitali-salvi (approved):

lisguo:
Seems like the checks are failing when fetching the AWS credentials...could there be a permissions issue?

lisguo (approved):

PR #43, opened 2024-05-21
Add ability to customize fluent bit config via values.yaml

sky333999:
*Description of changes:* Moving the fluent bit config to values.yaml to allow users to easily override it via helm and the Amazon CloudWatch Observability EKS add-on.

*Testing:* This change is essentially a no-op in terms of the rendered manifests. Confirmed the same by rendering the chart with & without my changes to ensure the content is the same.


charts/amazon-cloudwatch-observability/templates/linux/fluent-bit-configmap.yaml
```diff
@@ -8,195 +8,14 @@ metadata:
...
-        Time_Key            time
-        Time_Format         %Y-%m-%dT%H:%M:%S.%LZ
-
-    [PARSER]
-        Name                cwagent_firstline
-        Format              regex
-        Regex               (?<log>(?<="log":")\d{4}[\/-]\d{1,2}[\/-]\d{1,2}[ T]\d{2}:\d{2}:\d{2}(?!\.).*?)(?<!\\)".*(?<stream>(?<="stream":").*?)".*(?<time>\d{4}-\d{1,2}-\d{1,2}T\d{2}:\d{2}:\d{2}\.\w*).*(?=})
-        Time_Key            time
-        Time_Format         %Y-%m-%dT%H:%M:%S.%LZ
+    {{- .Values.containerLogs.fluentBit.config.customParsers  | nindent 4 }}
```
lisguo, line 16: is there a reason we split up the parsers and the fluentbit conf specifically? Seems like there's a lot of in-between configs too such as application-log, host-log, etc.

lisguo (approved):

charts/amazon-cloudwatch-observability/templates/linux/fluent-bit-configmap.yaml
sky333999: I tried to keep it close to the actual fluent-bit helm chart so that it looks familiar for users already using that.
https://github.com/fluent/helm-charts/blob/main/charts/fluent-bit/values.yaml#L367

Hence Ive exposed the service, customParsers and extraFiles as top level attributes. The application-log, host-log etc are specific to the way we have prescribed these files in our public docs in the past. We typically dont have root level inputs, filters and outputs in the main fluent-bit.conf and always take them in as additional/extra files.

mitali-salvi (approved):

PR #65, opened 2024-07-17
Allow configurable resource requests and limits through command arguments.

musa-asad:
*Description of changes:*
Allow configurable resource limits through command arguments so operator can take these values and use them for init containers.

*Testing:*
See https://github.com/aws/amazon-cloudwatch-agent-operator/pull/196.

1. With these local helm chart changes, I ran `helm upgrade --install amazon-cloudwatch-observability helm-charts/charts/amazon-cloudwatch-observability --values helm-charts/charts/amazon-cloudwatch-observability/values.yaml --set clusterName=cloudwatchagent-operator --set region=us-west-2`.
2. Then, ran `kubectl get pods`.
3. Then, ran `kubectl describe pod amazon-cloudwatch-observability-controller-manager-####`.

As we can see, it has the new operator image and command arguments.
Screenshot 2024-08-09 at 3 53 26 PM

### DIFF
--- a/charts/amazon-cloudwatch-observability/values.yaml
+++ b/charts/amazon-cloudwatch-observability/values.yaml
@@ -415,22 +415,22 @@ manager:
       tag: v1.32.3
       resources:
         requests:
-          memory: "50m"
-          cpu: "64Mi"
+          memory: "64Mi"
+          cpu: "50m"
         limits:
-          memory: "500m"
-          cpu: "64Mi"
+          memory: "64Mi"
+          cpu: "500m"
     python:
       repositoryDomain: public.ecr.aws/aws-observability
       repository: adot-autoinstrumentation-python
       tag: v0.3.0
       resources:
         requests:
-          memory: "50m"
-          cpu: "32Mi"
+          memory: "32Mi"
+          cpu: "50m"
         limits:
-          memory: "500m"
-          cpu: "32Mi"
+          memory: "32Mi"
+          cpu: "500m"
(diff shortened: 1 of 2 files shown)


charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -446,6 +446,13 @@ manager:
   service:
     name:
 
+## Default resource limits for init containers.
+initContainers:
+  resources:
+    limits:
+      cpu: 100m
+      memory: 64Mi
```
musa-asad, line 454: This is a placeholder default value, subject to change.

lisguo (approved):

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -410,14 +410,35 @@ manager:
       repositoryDomain: public.ecr.aws/aws-observability
       repository: adot-autoinstrumentation-java
       tag: v1.32.3
+      resources:
```
sky333999, line 413: These are not related to the image.

musa-asad: To assign different resource requests and limits per image, is that okay?

Paramadon (approved):

charts/amazon-cloudwatch-observability/values.yaml
musa-asad: Being addressed in https://github.com/aws-observability/helm-charts/pull/80.

Issue #54, opened 2024-06-25
should be able to set priorityClassName for fluent


brianmay:
My observation is unless I can set `priorityClassName: system-node-critical` for the fluent-bit daemon set, the cluster can end up in a broken state where it can't start fluent-bit on a node because there are not enough resources on a given node, but it will not kick off any of the pods that are consuming the resources either.

By setting this priority, it will kick off some of the less important pods, which will be rescheduled on other nodes, and fluent-bit is allowed to run.

But this chart doesn't appear to allow setting that value.

arenstar:
Second that - this chart requires priorityClassName

prem-pdmi:
When this will be released?

PR #70, opened 2024-07-25
feat(fluent-bit): set system-node-critical as a default priorityClass

m00lecule:
Issues:
#54 

In Kubernetes there is a good practice to specify a `system-node-critical` priority class for monitoring agents running as daemonsets.

ref: https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/

### DIFF
--- a/charts/amazon-cloudwatch-observability/values.yaml
+++ b/charts/amazon-cloudwatch-observability/values.yaml
@@ -50,7 +50,7 @@ containerLogs:
       requests:
         cpu: 50m
         memory: 25Mi
-    priorityClassName: system-node-critical
+    priorityClassName: ""
     config:
       service: |
         [SERVICE]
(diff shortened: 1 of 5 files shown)


charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -50,6 +50,7 @@ containerLogs:
       requests:
         cpu: 50m
         memory: 25Mi
+    priorityClassName: system-node-critical
```
lisguo, line 53: Can we set this to be empty by default? This will preserve existing behavior and will be opt in. Similar to [upstream otel operator](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/values.yaml#L313)

markandersontrocme:
@m00lecule we should also add the priorityClass to the cloudwatch-agent? https://github.com/aws-observability/helm-charts/blob/main/charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-daemonset.yaml#L23

markandersontrocme:
> @markandersontrocme I believe so, the general principle is to ensure that monitoring agents are running on all EKS nodes

I fee like it would make sense to add that to this PR 🙏

voidlily:
I've been manually working around this by setting the priorityClassName on the deployments/daemonsets, but that obviously doesn't scale well for version updates or deploying to multiple clusters. This PR would be useful to me as well (also adding `system-cluster-critical`, possibly `system-node-critical` since it's a daemonset) for the cloudwatch agent.

lisguo (approved):

sky333999 (approved):

PR #78, opened 2024-08-08
Allow Annotations on ServiceAccounts and Conditional Creation

hflobao:
*Description of changes:*

For the operator ServiceAccount, the configuration for 'create' and 'annotations' were present in the values file, but not evaluated in the template.

For the agent ServiceAccount, added the 'annotations' configuration in both the template and the values file.


miikatolonen:
Confirmed that this fix is working. Could this be reviewed and merged? We do not want to maintain custom charts. 

 @movence @mitali-salvi @lisguo @sky333999

PR #86, opened 2024-08-22
Add support for configurable modes for cloudwatch agent CR to support Prometheus monitoring with Target Allocator

mitali-salvi:
*Description of changes:*
1. Add support for configuring multiple agent workloads
2. Add support for modes (daemon-set, deployment, statefulset) for the cloudwatch-agent custom resource
3. Add support for prometheus monitoring via cloudwatch-agent
4. Add support for target allocator for prometheus monitoring
Add support for prometheusCR monitoring

*Testing*  - manually tested on cluster with the following config
```
agents:
  - name: prometheus-cloudwatch-agent
    mode: statefulset
    prometheus:
      config:
        global:
          scrape_interval: 60s
        scrape_configs:
          - job_name: 'cadvisor'
            scrape_interval: 15s
            static_configs:
              - targets: [ 'cadvisor:8080' ]
      targetAllocator:
        enabled: true
```


charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -529,6 +529,12 @@ admissionWebhooks:
   ## Secret labels
   secretLabels: { }
 
+## Agent workloads list
+agentWorkloads:
+  - name: cloudwatch-agent-with-container-insights-and-application-signals
+    ## Supported modes are daemonset, statefulset and deployment
```
lisguo, line 535: any reason to not just expose config here? I know there is a default config, but why not just put the default config directly in the `agentWorkloads` section to make it explicit?

lisguo:
Helm chart integ tests should still pass right? Is the agent running as expected?

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -529,6 +529,12 @@ admissionWebhooks:
   ## Secret labels
   secretLabels: { }
 
+## Agent workloads list
+agentWorkloads:
```
lisguo, line 533: is `agentWorkloads` going to be the new `agent` in the eks addon advanced config? Would we break customers if we change that?

charts/amazon-cloudwatch-observability/templates/target-allocator-clusterrolebinding.yaml
```diff
@@ -0,0 +1,21 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  labels:
+    {{- include "amazon-cloudwatch-observability.labels" $ | nindent 4 }}
+  name: "cloudwatch-agent-target-allocator-rolebinding"
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: "cloudwatch-agent-target-allocator"
```
okankoAMZ, line 13:
```suggestion
  name: "cloudwatch-agent-target-allocator-role"
```
The role is named `cloudwatch-agent-target-allocator-role` it can be seen [here](https://github.com/aws-observability/helm-charts/blob/b83c1dfbac899a5dc5e07e7ce195dc0e2d3c274f/charts/amazon-cloudwatch-observability/templates/target-allocator-clusterrole.yaml#L9)

okankoAMZ (commented):
Could you update the desc, with an example on how to have multiple workloads

okankoAMZ (approved):
Minor comments

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -545,8 +545,14 @@ admissionWebhooks:
   ## Secret labels
   secretLabels: { }
 
+## Agent workloads list
```
sky333999, line 548:
```suggestion
## List of AmazonCloudWatchAgent workloads to install & manage, each representing an independent installation of the AmazonCloudWatchAgent CustomResource. Each entry in this list uses the schema & the defaults from $.agent, so only provide any overrides here.
```

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -545,8 +545,14 @@ admissionWebhooks:
...
+agents:
+  - name: cloudwatch-agent
+
 agent:
   name:
+  mode: # represents the mode the cloudwatch-agent will run in (deployment, daemonset or statefulset)
```
sky333999, line 554: Shouldnt this default to daemonset?

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -545,8 +545,14 @@ admissionWebhooks:
...
+  replicas: # The total number non-terminated pods targeted by this cloudwatch-agent's deployment or statefulSet.
```
sky333999, line 555: Maybe set this to -1 as the default to make it obvious its not applicable by default and anything using it will fail unless overriden.

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -602,6 +608,37 @@ agent:
         }
       }
     }
+  windowsDefaultConfig:
```
sky333999, line 611: This content actually now looks identical to the linux default above.
It might actually be cleaner to now have `cloudwatch-agent-windows` in the `agents` list above and drop this windowsDefaultConfig field here. We can instead add an `os` field thats either `linux` or `windows` and spin up all other resources based on that flag.

So `agents` would look like:
```
agents:
- name: cloudwatch-agent
- name: cloudwatch-agent-windows
  os: windows
```

sky333999: Its fine if we do this in a separate PR.

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -602,6 +608,37 @@ agent:
...
+      },
+      "traces": {
+        "traces_collected": {
+          "application_signals": { }
+        }
+      }
+    }
+  prometheus:
+    config:
+    targetAllocator:
```
sky333999, line 629: i think we'll have more TLS related stuff here, but those can come slightly later

charts/amazon-cloudwatch-observability/templates/operator-clusterrole.yaml
```diff
@@ -51,3 +51,6 @@ rules:
 - apiGroups: [ "route.openshift.io" ]
   resources: [ "routes", "routes/custom-host" ]
   verbs: [ "create","delete","get","list","patch","update","watch" ]
+- apiGroups: [ "policy" ]
+  resources: [ "poddisruptionbudgets" ]
```
sky333999, line 55: Is the latest operator creating PDBs for the deployment mode?

okankoAMZ (commented):
how are passing the target allocator definitions of the agent to the operator so that, that version can be deployed

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -602,6 +608,37 @@ agent:
...
+      enabled: false
+      image:
+        repository: cloudwatch-agent-target-allocator
```
okankoAMZ, line 632: I think this section should be under operator since the TA image needs to be given to operator

okankoAMZ (changes requested):
We are passing the TA image to operator

musa-asad:
https://github.com/aws-observability/helm-charts/blob/target-allocator/charts/amazon-cloudwatch-observability/templates/operator-deployment.yaml#L35-L36

We should include between the above lines so that the defined target allocator image is used in the operator.
```yaml
"--target-allocator-image={{ template "target-allocator.modify-image" . }}"
```

charts/amazon-cloudwatch-observability/values.yaml
okankoAMZ: Nevermind we do not need this due this line on this code: https://github.com/aws/amazon-cloudwatch-agent-operator/blob/a8771928399017e02f7507f3e22efe9d228d6180/internal/manifests/targetallocator/container.go#L18

okankoAMZ (approved):

sky333999:
Approving since its to a dev branch. Comments will be addressed in a follow up.

sky333999 (approved):

PR #97, opened 2024-09-10
feat: Support runtime-metrics

adebayor123:
*Issue #, if available:*

*Description of changes:*
As part of supporting runtime metrics for Application Signals, we need a mechanism to allow customers to disable runtime-metrics for Addons. This PR creates a new field under `manager`, `autoInstrumentationApplicationSignals` that will be used in the cloudwatch-agent-operator to disable runtime-metrics.


bjrara (approved):

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -422,6 +422,13 @@ manager:
       repositoryDomain: public.ecr.aws/aws-observability
       repository: adot-autoinstrumentation-node
       tag: v0.2.0
+  autoInstrumentationApplicationSignals:
```
sky333999, line 425: Given the other `autoInstrumentation` fields in here are all also for ApplicationSignals, should we just make this: 

```
autoInstrumentationRuntimeMetrics:
  java:
    enabled: "true"
```

jefchien: Or if we want to keep it more generic to support future flags `autoInstrumentationConfiguration`.

bjrara: Changed it to `autoInstrumentationConfiguration`.

jefchien (approved):

sky333999 (approved):

PR #120, opened 2024-10-28
Add agent server port for vending entity to fluent bit

lisguo:
*Issue #, if available:*

*Description of changes:*
- Adds certificates for agent server port that exposes service name per instrumented pod 
- Adds fluent bit configuration to retrieve service name from the agent server port to populate entity to CW Logs


zhihonl (approved):

okankoAMZ (commented):
When Agent is set to be stateful said and has multiple replicas are they all using the same certs?

okankoAMZ (approved):

zhihonl (approved):

PR #121, opened 2024-10-30
[TA] TLS Cert Generation for Multiple Target Allocator Services

okankoAMZ:
*Issue #, if available:*
Currently, we cannot support TLS for target allocator because TLS certs are not accepting TA service as a valid dns. This PR dynamically generates the TA service dns' for each agent that uses TA.
*Description of changes:*
* Updated Cert Manager to generate TA dns.
* Updated key generation to generate TA dns.

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/certmanager.yaml
+++ b/charts/amazon-cloudwatch-observability/templates/certmanager.yaml
@@ -55,6 +55,9 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   dnsNames:
+    {{- range $i, $customAgent := .Values.agents }}
+    - {{( printf "%s-target-allocator-service" $customAgent.name )}}
+    {{- end }}
     - "dcgm-exporter-service"
     - "dcgm-exporter-service.amazon-cloudwatch.svc"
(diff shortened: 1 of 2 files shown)


musa-asad:
Can you show that this works?

musa-asad (approved):

charts/amazon-cloudwatch-observability/templates/certmanager.yaml
```diff
@@ -55,6 +55,9 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   dnsNames:
+    {{- range $i, $customAgent := .Values.agents }}
+    - {{( printf "%s-target-allocator-service" $customAgent.name )}}
```
mitali-salvi, line 59: Can we use a single DNS name for all the agent workloads ? Why do we need to create separate DNS names for every agent workload ?

mitali-salvi: Spoke to @okankoAMZ offline, the helm-chart will support multiple target-allocators and hence will use DNS names for every workload

mitali-salvi (approved):

PR #145, opened 2024-12-18
Pass TA Image to Operator Deployment

musa-asad:
*Description of changes:*
- Pass `target-allocator` image to `operator-deployment.yaml`

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/operator-deployment.yaml
+++ b/charts/amazon-cloudwatch-observability/templates/operator-deployment.yaml
@@ -32,6 +32,7 @@ spec:
         - "--auto-instrumentation-python-image={{ template "auto-instrumentation-python.image" . }}"
         - "--auto-instrumentation-dotnet-image={{ template "auto-instrumentation-dotnet.image" . }}"
         - "--auto-instrumentation-nodejs-image={{ template "auto-instrumentation-nodejs.image" . }}"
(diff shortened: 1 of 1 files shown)


charts/amazon-cloudwatch-observability/templates/operator-deployment.yaml
```diff
@@ -32,6 +32,7 @@ spec:
         - "--auto-instrumentation-python-image={{ template "auto-instrumentation-python.image" . }}"
         - "--auto-instrumentation-dotnet-image={{ template "auto-instrumentation-dotnet.image" . }}"
         - "--auto-instrumentation-nodejs-image={{ template "auto-instrumentation-nodejs.image" . }}"
+        - "--target-allocator-image={{ template "target-allocator.modify-image" }}"
```
movence, line 35: don't we need `imageDomain` like other images for `target-allocator.modify-image` in template helper?

musa-asad: I believe it's present, but in its own block below.

https://github.com/aws-observability/helm-charts/blob/73382ddb0d7626973fd032c1356026ab8b667774/charts/amazon-cloudwatch-observability/templates/_helpers.tpl#L153-L159

movence (approved):

PR #152, opened 2024-12-23
Release v3.0.0

sky333999:
*Description of changes:* Release v3.0.0


RELEASE_NOTES
```diff
@@ -1,3 +1,21 @@
+========================================================================
+amazon-cloudwatch-observability v3.0.0 (2024-12-23)
+========================================================================
+New Features:
+* Add support for installing multiple AmazonCloudWatchAgent workloads (#126)
```
musa-asad, line 5:
```suggestion
* Add support for installing multiple CloudWatchAgent workloads (#126)
```
For consistency.

musa-asad (approved):

RELEASE_NOTES
```diff
@@ -1,3 +1,21 @@
...
+* Introduce support for Prometheus scraping with the CloudWatchAgent and TargetAllocator (#126)
+* Set default PriorityClass as system-node-critical for CloudWatchAgent and FluentBit daemonsets (#137)
+
+Enhancements:
+* Unify schema for Application Signals metrics across platforms
+* Increment ADOT JAVA SDK to v1.32.6 for Application Signals
```
mitali-salvi, line 11: Was this done as part of a separate PR ?

mitali-salvi (approved):

RELEASE_NOTES
sky333999: The custom resource is actually called AmazonCloudWatchAgent which is why I used that for this line. Its not referring to the agent pods, but the custom resource instead.

musa-asad: Oh, right. Makes sense.

PR #154, opened 2025-01-06
Support per-component tolerations (+ manager)

loganmzz:
Should partially fix https://github.com/aws/containers-roadmap/issues/2515

*Issue #, if available:*

https://github.com/aws/containers-roadmap/issues/2515

*Description of changes:*

1. Let's support per-component tolerations
2. Add tolerations for `manager`

*Behavior*

* Root tolerations is applied by default
* Manager tolerations is disabled by default (`[]` value, see below)
* Each component may override tolerations
* Tolerations can be disabled by using empty list `[]` (root or specific)
* If specific tolerations is `null` (`nil` in Go/Template), root ones are used.

---

### DIFF
--- a/charts/amazon-cloudwatch-observability/values.yaml
+++ b/charts/amazon-cloudwatch-observability/values.yaml
@@ -186,6 +186,8 @@ containerLogs:
       requests:
         cpu: 50m
         memory: 25Mi
+    ## Override root tolerations
+    # tolerations: []
     priorityClassName: system-node-critical
     config:
       service: |
@@ -1263,6 +1265,7 @@ manager:
     requests:
       cpu: 100m
       memory: 64Mi
+  tolerations: []
   ## Adds additional environment variables
   ## e.g ENV_VAR: env_value
   env:
@@ -1340,6 +1343,8 @@ agent:
     limits:
       memory: 512Mi
       cpu: 500m
+  ## Override root tolerations
+  # tolerations: []
   ## TLS Certificate Option 1: Use Helm to automatically generate self-signed certificate.
   ## autoGenerateCert must be enabled. This is the default option.
   ## If true, Helm will automatically create a self-signed cert and secret for you.
@@ -1411,6 +1416,8 @@ dcgmExporter:
     limits:
       cpu: 500m
       memory: 500Mi
+  ## Override root tolerations
+  # tolerations: []
   configmap: dcgm-exporter-config-map
   arguments:
     - --web-config-file=/etc/dcgm-exporter/web-config.yaml
@@ -1436,6 +1443,8 @@ neuronMonitor:
     requests:
       cpu: 256m
       memory: 128Mi
+  ## Override root tolerations
+  # tolerations: []
   configmap: neuron-monitor-config-map
   service:
     enable: true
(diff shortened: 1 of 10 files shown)


charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1272,6 +1272,9 @@ manager:
   podLabels: {}
   service:
     name:
+  # When true, root tolerations aren't added to this one.
+  tolerationsReplaced: false
```
sky333999, line 1276: It might not be very intuitive to set another boolean just to express that the tolerations at the specific component level should be respected.

Could we instead have the `tolerations` commented out at the specific component level, such as:
```
# tolerations: []  # Uncomment this line to override root-level tolerations.
```

Once you merge from latest main, you will notice the root tolerations have been defined & only apply to daemonsets and do NOT apply to the controller manager deployment. So just for controller manager alone, we truly do want to explicitly set `tolerations: []` and let it override the root level ones to maintain compatibility i.e. not comment that line.

We can then probably have logic that does something like:
```
{{- if hasKey .Values.manager "tolerations" -}}
  {{- toYaml .Values.manager.tolerations -}}
{{- else -}}
  {{- toYaml .Values.tolerations -}}
{{- end -}}
{{- end -}}
```

Could probably put this in a shared method in _helpers.tpl and call it for the all components.

The components to cover would be:
* manager:
* containerLogs.fluentBit
* agent
* dcgmExporter
* neuronMonitor

loganmzz: Hi @sky333999,

Thank you for the feedback!

The PR has been designed to be conservative to avoid unexpected impact for existing deployment. I will check new code changes and update my PR accordingly.

sky333999: Thats great! Thank you so much for the changes. Looks good to me.

sky333999:
Thanks for the updates @loganmzz . Would you mind adding some testing information to the description so we have that captured before I merge this in?
Some manual test suggestions if you could please try out locally:
* Render the chart prior to and after your change to make sure the rendered manifests are identical since your change should be a no-op
* Override the deployment tolerations to something other than [] and ensure its reflected correctly in the rendered manifests
* Override one of the daemonset tolerations to [] or something eles to ensure its reflected correctly in the rendered manifests

charts/amazon-cloudwatch-observability/values.yaml
sky333999: Now that I think about it a bit more, you could probably get away with something like:
```
  {{- with (.Values.agent.tolerations | default .Values.tolerations ) }}
  tolerations: {{- toYaml . | nindent 2}}
  {{- end }}
```

Im debating if this is just simpler for each of the components instead of the helper function. Im fine either way.

loganmzz: Sorry I miss your two last review feedbacks.

Because `nil` and empty are handled differently `default` function won't do what is expected.

loganmzz: @sky333999 Any plan to merge this?

sky333999 (approved):

charts/amazon-cloudwatch-observability/values.yaml
sky333999: Thanks for following up. Ive approved the PR. Will get another admin to approve this next week to see if we can get it merged.

mitali-salvi (approved):

sky333999:
Hey @loganmzz - thanks again for your contribution.
We currently have a limitation of not being able to run our integ test checks on forks. Additionally, we built on top of your change to also allow configuring node selectors & affinity at a component level.
We ensured to cherry pick your commit to retain your changes & author history as part of https://github.com/aws-observability/helm-charts/pull/205.

Think we can close this PR.

PR #163, opened 2025-01-20
Set up mTLS resources for the Amazon CloudWatch Agent client

musa-asad:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/certmanager.yaml
+++ b/charts/amazon-cloudwatch-observability/templates/certmanager.yaml
@@ -44,7 +44,7 @@ spec:
 {{- end }}
 {{- end }}
 
-  {{- if ( .Values.agent.certManager.enabled) }}
+{{- if ( .Values.agent.certManager.enabled) }}
 ---
 apiVersion: cert-manager.io/v1
 kind: Certificate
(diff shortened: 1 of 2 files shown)


charts/amazon-cloudwatch-observability/templates/certmanager.yaml
```diff
@@ -100,7 +100,25 @@ spec:
...
+    - digital signature
+    - key encipherment
+    - cert sign
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  labels:
+    {{- include "amazon-cloudwatch-observability.labels" . | nindent 4 }}
+  name: "amazon-cloudwatch-observability-agent-outbound-cert"
```
lisguo, line 114: is there a reason why we can't use `"amazon-cloudwatch-observability-agent-cert"` ? 

I see that we specify the target allocator service as a dns name we can make calls to? 

Are we using this outbound cert specifically for target allocator?

musa-asad: This is specifically the client cert and key generated for the agent (when behaving as a client), we can't use `amazon-cloudwatch-observability-agent-cert` since that is where the server cert and key are generated.

To clarify, we do use `amazon-cloudwatch-observability-agent-cert` to generate the server-side cert and key for the target allocator, but we need `amazon-cloudwatch-observability-agent-outbound-cert` to generate the client-side certa nd key for the agent.

okankoAMZ: Could you elaborate on what you meant by
>I avoided doing so due to separation of concerns.

musa-asad: The `server-cert` is used to generate server cert and keys when the CloudWatch Agent operates as a server for FluentBit. 

The `outbound-cert` is used to generate client cert and keys when the CloudWatch Agent operates as a client for the Target Allocator.

We are able to use the `server-cert` to generate _client_ cert and keys as well when the CloudWatch Agent operates as a client, but I wanted to separate that functionality to the `outbound-cert`.

musa-asad: Updated the naming from `outbound-cert` to `ta-client-cert`.

okankoAMZ (dismissed):

PR #188, opened 2025-03-27
Implement exponential backoff by configuring kube-client env vars

musa-asad:
*Description of changes:*
The CloudWatch Agent overloads the API server due to there not being an exponential back-off strategy implemented in the Kubernetes client when there's a timeout issue, as described in https://github.com/cilium/cilium/issues/36525#issuecomment-2541833105.

This change configures environmental variables that enables exponential back-off for timeout issues.

### DIFF
--- a/charts/amazon-cloudwatch-observability/values.yaml
+++ b/charts/amazon-cloudwatch-observability/values.yaml
@@ -1451,3 +1451,7 @@ neuronMonitor:
         - SYS_ADMIN
   serviceAccount:
     name: # override exporter service account name
+k8sClientExponentialBackoff:
+  enabled: true
+  backoffBaseSeconds: 1
+  backoffMaxDurationSeconds: 120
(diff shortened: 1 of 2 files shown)


charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1451,3 +1451,7 @@ neuronMonitor:
         - SYS_ADMIN
   serviceAccount:
     name: # override exporter service account name
+k8sClientExponentialBackoff:
```
sky333999, line 1454: Defining this at the global level sets an expectation that this will be respected for all k8s clients across every component deployed by this chart, including the operator, agent, fluentbit and anything else - but we arent doing that.

So making this agent specific might be better - but then why is this being defined as a helm override for the agent and not part of the agent json config like everything else.

We need more discussions on where these belong.

lisguo:
We should test this to ensure that it does actually reduce the number of control plane api calls

charts/amazon-cloudwatch-observability/values.yaml
musa-asad: Yeah, completely agree.

PR #194, opened 2025-04-15
Quote the cluster name environment variable to ensure it remains a string

### LINKED ISSUE #193: Some cluster names can cause helm chart deployment to fail
The amazon-cloudwatch-observability helm chart takes a cluster name as required input. This is used as an environment variable in a cloudwatch custom resource here: https://github.com/aws-observability/helm-charts/blob/de84bbde3ac9de6898459ccfdde4a7aca4065f6b/charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml#L236

This has an issue, where if the cluster name is something which yaml interprets as a non-string type, the deployment will fail due to type mismatch.

For example, a cluster name `yes`, `no`, `true`, `false`, `205930`, `1.2305`, etc will all cause this issue.

This is due to the fact that this cluster name is not surrounded by quotes in the helm template where it is used.

Whenever templating parameters, the helm chart should ensure to surround values which are expected to be strings with quotes for protection against this.

cheeseandcereal:
*Issue #, if available:*

Fixes #193

*Description of changes:*

This CR modifies the templating of the environment variables in regards to the user-provided cluster name, to surround it with quotes, to ensure that it remains a string regardless of the cluster name. This fixes issues where a cluster name may be something like `409238` which gets translated in yaml to a number, not a string without quotes.

*Testing:*
Numerical:
Screenshot 2025-04-16 at 3 27 49 PM
Screenshot 2025-04-16 at 3 28 04 PM

String:
Screenshot 2025-04-16 at 3 29 49 PM
Screenshot 2025-04-16 at 3 29 41 PM

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
+++ b/charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
@@ -233,7 +233,7 @@ spec:
     value: "True"
   {{ end }}
   - name: K8S_CLUSTER_NAME
-    value: {{ $.Values.clusterName }}
+    value: "{{ $.Values.clusterName }}"
   {{- dict "component" $agent "context" $ | include "amazon-cloudwatch-observability.common.tolerations" | nindent 2 }}
 ---
 {{- end }}
--- a/charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
+++ b/charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
@@ -233,7 +233,7 @@ spec:
     value: "True"
   {{ end }}
   - name: K8S_CLUSTER_NAME
-    value: "{{ $.Values.clusterName }}"
+    value: {{ $.Values.clusterName | quote }}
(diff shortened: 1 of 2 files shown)


charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
```diff
@@ -233,7 +233,7 @@ spec:
     value: "True"
   {{ end }}
   - name: K8S_CLUSTER_NAME
-    value: {{ $.Values.clusterName }}
+    value: "{{ $.Values.clusterName }}"
```
musa-asad, line 236: To keep it consistent with our other code.
```suggestion
    value: {{ $.Values.clusterName | quote }}
```

musa-asad (approved):

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
cheeseandcereal: Looks good. I saw it done both ways in the helm chart. But either way is just as good to me

musa-asad (approved):

musa-asad (approved):

PR #209, opened 2025-04-24
fix bugs around prometheus cr

gane5hvarma:
*Issue #, if available:*

*Description of changes:*
The helm charts have bugs around enabling prometheusCR. If prometheusCR is enabled true, it leads to this error.
```
Error: UPGRADE FAILED: cannot patch "cloudwatch-agent" with kind AmazonCloudWatchAgent: admission webhook "mamazoncloudwatchagent.kb.io" denied the request: json: cannot unmarshal bool into Go struct field AmazonCloudWatchAgentTargetAllocator.spec.targetAllocator.prometheusCR of type v1alpha1.AmazonCloudWatchAgentTargetAllocatorPrometheusCR
```

 The value is  passed as 
```
prometheusCR: true
```
Instead of this
```
prometheusCR:
  enabled: true
```
Also there is case bug in target allocator cluster role yaml


gane5hvarma:
@Paramadon @mitali-salvi  @lisguo can you take a look at this pr and do the needful.

movence (approved):

PR #219, opened 2025-06-24
upgrade neuron monitor version to 1.5.1

movence:
*Description of changes:*
Upgrade Neuron monitor version to 1.5.1

### DIFF
--- a/charts/amazon-cloudwatch-observability/values.yaml
+++ b/charts/amazon-cloudwatch-observability/values.yaml
@@ -1502,7 +1502,7 @@ neuronMonitor:
   name:
   image:
     repository: neuron-monitor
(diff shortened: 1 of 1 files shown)


movence:
Looks like block style isn't supported by the helm version being installed

```
Installed hashicorp/helm v3.0.2 (signed by HashiCorp)
```
`v3.0.2` only supports argument type so
```
provider "helm" {
  kubernetes {
    config_path = "${var.kube_dir}/config"
  }
}
should be changed to (with '=')
provider "helm" {
  kubernetes = {
    config_path = "${var.kube_dir}/config"
  }
}

TravisStark (approved):

PR #242, opened 2025-09-18
Add automatic rolling deployment & option to re-create admission webhook certs

duhminick:
*Issue #, if available:* NA

*Description of changes:*

**Certificate Management Enhancement:**
- Added a `recreate` flag for auto-generated certificates in the admission webhooks configuration
- This flag allows users to control whether existing certificates should be recreated during deployments
- Updated the helper templates and operator webhook configuration to support this new functionality
  - Will first check to see if autoGenerateCert is enabled, then check to see if there is an already existing cert with the recreate option set to false. If so, then it will retrieve the cert info. The ca.crt is not in the secret so we retrieve that from the caBundle from the admission webhooks to be re-used. Otherwise it will create new certs - the logic is taken from the operator-webhook.yaml file (copy & paste).
  - The operator-webhook.yaml file is updated to use the cert values from the helper
- Default value is set to `true` to maintain backward compatibility

**Operator Deployment Rolling:**
- Technique taken from: https://v3.helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments
- Introduced a `rolling` flag in the operator configuration to enable automatic deployment rolling
- When enabled, this forces a deployment spec change to trigger pod restarts without manual intervention
- Default value is set to `false` to preserve existing behavior

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/_helpers.tpl
+++ b/charts/amazon-cloudwatch-observability/templates/_helpers.tpl
@@ -342,3 +342,33 @@ Define the default service name
 {{- define "amazon-cloudwatch-observability.webhookServiceName" -}}
 {{- default (printf "%s-webhook-service" (include "amazon-cloudwatch-observability.name" .)) .Values.manager.service.name }}
 {{- end -}}
+
+{{/*
+Returns auto-generated certificate and CA for admission webhooks.
+*/}}
+{{- define "amazon-cloudwatch-observability.webhookCert" -}}
+{{- $tlsCrt := "" }}
+{{- $tlsKey := "" }}
+{{- $caCrt := "" }}
+{{- if .Values.admissionWebhooks.autoGenerateCert.enabled }}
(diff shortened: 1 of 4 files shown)


duhminick:
Reference PRs:
1. https://github.com/open-telemetry/opentelemetry-helm-charts/pull/936
2. https://github.com/open-telemetry/opentelemetry-helm-charts/pull/720

sky333999 (dismissed):

integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/deployment_rolling_enabled_test.go
```diff
@@ -0,0 +1,54 @@
...
+	assert.NotNil(t, actualData)
+
+	savedData, err := os.ReadFile(deploymentRollingEnabledPath)
+	assert.NoError(t, err)
+
+	assert.False(t, bytes.Equal(actualData, savedData))
+}
+
+func retrieveRollMe(t *testing.T, k8sClient util.K8sClient) []byte {
+	ds, err := k8sClient.ListDeployments("amazon-cloudwatch")
```
movence: use existing values from `common.go`

movence (approved):

PR #250, opened 2025-10-16
release v5.0.0

sky333999:
*Description of changes:* Release v5.0.0

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,4 +1,18 @@
 =======================================================================
+amazon-cloudwatch-observability v5.0.0 (2025-10-16)
+=======================================================================
+Enhancements:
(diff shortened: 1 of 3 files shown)


ejether:
👋  Hey there! This has some useful changes I'm interested in. Is there any estimate on when it may get cut?  🙇

sky333999:
Hey @ejether, v5.0.0 is on a bit of a slow schedule - so we are letting the corresponding EKS add-on for this helm chart to out out in waves before we can release it here.

But most of the changes are still going out in the v4.x.x series in parallel. So we recently released v4.7.0 for instance.
Can you check if the change you are interested in has already been released? If not, we can bundle that into an upcoming v4.x.x release so its available sooner.

PR #254, opened 2025-12-01
Adding top level useDualstackEndpoint configuration

Paramadon:
## Issue #, if available
Currently customers can not use our default configuration to have agent and fluent bit use dualstack endpoints. We want to add a top level field to allow using dualstack endpoints.

Passing tests: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/19836119811
## Description of changes:

If customer sets the new top level `useDualstackEndpoint` field to true, we'll add `use_dualstack_endpoint` to true in the agent configuration so that it uses dualstack endpoints(this is the current way to enable cloudwatch agent to use dualstack endpoints.) 

As for fluent bit, we'll update `endpoints/sts_endpoint` for fluent bit configuration (under `[OUTPUT]` section) to use dualstack endpoints and then have fluent bit prefer ipv6 dns using net.dns.prefer_ipv6 true.  (Note setting `useDualstackEndpoint` to `true` on an ipv4 instance would **NOT** break the cluster as dualstack endpoints also work on ipv4 instances and net dns prefer ipv6 does fall back to ipv4. (Applied helm on ipv4 instance and set `useDualstackEndpoint`, and it works.

Also we do aren't replacing endpoints if there is already an endpoint field (for example adc regions). ADC regions are out of scope of this change. 

## Testing
All testing was done on IPv6 EKS cluster, we also tested that dualstack endpoints work on K8s so we can make this helm chart change. 

Below are logs from fluent bit and the agent configuration showcasing what happens if we set the top level field `useDualstackEndpoint` to true or false.

### Helm upgrade command:
```
helm upgrade --install amazon-cloudwatch-observability \
  charts/amazon-cloudwatch-observability \
  -n amazon-cloudwatch \
  -f charts/amazon-cloudwatch-observability/values.yaml \
  --set clusterName=<cluster-name> \
  --set region=us-west-2 \
  --set useDualstackEndpoint=true
  ```
  
 ### Resulting agent configuration

```
 config: '{"agent":{"region":"us-west-2","use_dualstack_endpoint":true},"logs":{"metrics_collected":{"application_signals":{"hosted_in":"Ipv6-auto-mode-6n"},"kubernetes":{"cluster_name":"Ipv6-auto-mode-6n","enhanced_container_insights":true}}},"traces":{"traces_collected":{"application_signals":{}}}}
```

As you can setting useDualstackEndpoint field to `true` sets agent configuration to use dualstack endpoint.

 ### Resulting fluent bit config
 ```
[SERVICE]
  net.dns.prefer_ipv6       true
  Flush                     5
  Grace                     30
  Log_Level                 error
  Daemon                    off
  Parsers_File              parsers.conf
  storage.path              /var/fluent-bit/state/flb-storage/
  storage.sync              normal
  storage.checksum          off
  storage.backlog.mem_limit 5M
  
[INPUT]
 Name tail
 Path /var/log/containers/*.log
 Parser docker

[OUTPUT]
  Name                cloudwatch_logs
  Match               application.*
  region              ${AWS_REGION}
  endpoint            logs.${AWS_REGION}.api.aws
  sts_endpoint        sts.${AWS_REGION}.api.aws
  log_group_name      /aws/containerinsights/${CLUSTER_NAME}/application
  log_stream_prefix   ${HOST_NAME}-
  auto_create_group   true
  extra_user_agent    container-insights


 ```

 ### Resulting fluent bit logs
As for Fluent bit, here are some log output showcasing successfully sending logs using dualstack endpoints:

Screenshot 2025-12-01 at 3 34 40 PM

### Helm upgrade command:
```
helm upgrade --install amazon-cloudwatch-observability \
  charts/amazon-cloudwatch-observability \
  -n amazon-cloudwatch \
  -f charts/amazon-cloudwatch-observability/values.yaml \
  --set clusterName=<cluster-name> \
  --set region=us-west-2 \
  --set useDualstackEndpoint=false
  ```

### Resulting agent configuration
```
config: '{"agent":{"region":"us-west-2"},"logs":{"metrics_collected":{"application_signals":{"hosted_in":"Ipv6-auto-mode-6n"},"kubernetes":{"cluster_name":"Ipv6-auto-mode-6n","enhanced_container_insights":true}}},"traces":{"traces_collected":{"application_signals":{}}}}
```

As you can see seeting `useDualstackEndpoint` to `false` does not set the `use_dualstack_endpoint` to true nor adds it to the agent configuration. 

 ### Resulting fluent bit config
 ```
 [SERVICE]
  Flush                     5
  Grace                     30
  Log_Level                 error
  Daemon                    off
  Parsers_File              parsers.conf
  storage.path              /var/fluent-bit/state/flb-storage/
  storage.sync              normal
  storage.checksum          off
  storage.backlog.mem_limit 5M
  
[INPUT]
 Name tail
 Path /var/log/containers/*.log
 Parser docker


[OUTPUT]
  Name                cloudwatch_logs
  Match               application.*
  region              ${AWS_REGION}
  log_group_name      /aws/containerinsights/${CLUSTER_NAME}/application
  log_stream_prefix   ${HOST_NAME}-
  auto_create_group   true
  extra_user_agent    container-insights

 ```

### Resulting fluent bit logs
As for Fluent bit, here are some log output showcasing successfully sending logs using regular ipv4 endpoints.

Screenshot 2025-12-01 at 3 41 44 PM

### Test to ensure we don't override endpoint if it is already given

In the fluent bit configuration I added endpoint `logs.us-east-1.api.aws` to see if the endpoint is overwritten with `useDualstackEndpoint` is true and it isn't overridden and the fluent bit configuration does take precedence making sure that we don't override adc region endpoints. 

Screenshot 2025-12-02 at 3 08 03 PM

### Added scenario tests
Added minikube-based integration tests to validate the useDualstackEndpoint configuration:

What each test validates:

`default_test.go`: Confirms FluentBit configs don't contain dualstack endpoints or IPv6 preference when disabled

`dualstack_endpoint_enabled_test.go`: Confirms FluentBit `OUTPUT` sections have endpoint: `logs.${AWS_REGION}.api.aws` and sts_endpoint: `sts.${AWS_REGION}.api.aws`, `SERVICE` section has `net.dns.prefer_ipv6` true, and CloudWatch Agent has `use_dualstack_endpoint`: true

`dualstack_endpoint_with_custom_endpoint_test.go`: Confirms custom endpoints in `application-log.conf` are NOT overwritten, while other configs still get dualstack endpoints

### DIFF
--- a/integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/dualstack_endpoint_with_custom_endpoint_test.go
+++ b/integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/dualstack_endpoint_with_custom_endpoint_test.go
@@ -4,14 +4,11 @@
 package scenarios
 
 import (
-	"context"
-	"encoding/json"
 	"testing"
 
 	"github.com/aws-observability/helm-charts/integration-tests/amazon-cloudwatch-observability/util"
 	"github.com/aws-observability/helm-charts/integration-tests/amazon-cloudwatch-observability/validations/minikube"
 	"github.com/stretchr/testify/assert"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestDualstackEndpointWithCustomEndpoint(t *testing.T) {
@@ -39,7 +36,7 @@ func TestDualstackEndpointWithCustomEndpoint(t *testing.T) {
 	})
 
 	t.Run("CloudWatchAgentStillHasDualstack", func(t *testing.T) {
-		validateCloudWatchAgentStillHasDualstack(t, k8sClient)
+		validateCloudWatchAgentDualstackEndpoint(t, k8sClient)
 	})
 
 	t.Run("FluentBitIPv6PreferenceStillSet", func(t *testing.T) {
@@ -74,44 +71,6 @@ func validateOtherConfigsStillGetDualstack(t *testing.T, k8sClient *util.K8sClie
 		assert.Contains(t, conf, "logs.${AWS_REGION}.api.aws", "%s should contain dualstack logs endpoint since no custom endpoint is set", configName)
 		assert.Contains(t, conf, "sts.${AWS_REGION}.api.aws", "%s should contain dualstack sts endpoint since no custom endpoint is set", configName)
 	}
-}
-
-func validateCloudWatchAgentStillHasDualstack(t *testing.T, k8sClient *util.K8sClient) {
-	dynamicClient, err := k8sClient.GetDynamicClient()
-	assert.NoError(t, err)
-
-	agentList, err := dynamicClient.Resource(getAmazonCloudWatchAgentGVR()).
-		Namespace(minikube.Namespace).
-		List(context.Background(), metav1.ListOptions{})
-	assert.NoError(t, err)
-	assert.NotEmpty(t, agentList.Items, "No CloudWatch Agent resources found")
-
-	foundDualstackConfig := false
-	for _, agent := range agentList.Items {
-		spec, found := agent.Object["spec"].(map[string]any)
-		if !found {
-			continue
-		}
-
-		configStr, found := spec["config"].(string)
-		if !found {
-			continue
-		}
-
-		var config map[string]any
-		if err := json.Unmarshal([]byte(configStr), &config); err != nil {
-			continue
-		}
-
-		if agentConfig, ok := config["agent"].(map[string]any); ok {
-			if useDualstack, ok := agentConfig["use_dualstack_endpoint"].(bool); ok && useDualstack {
-				foundDualstackConfig = true
-				break
-			}
-		}
-	}
-
-	assert.True(t, foundDualstackConfig, "use_dualstack_endpoint should still be true in CloudWatch Agent config even with custom FluentBit endpoint")
 }
 
 func validateFluentBitIPv6PreferenceStillSet(t *testing.T, k8sClient *util.K8sClient) {


charts/amazon-cloudwatch-observability/templates/_helpers.tpl
```diff
@@ -210,6 +217,17 @@ Get the current recommended fluent-bit image for a region
 {{- printf "%s/%s:%s" $imageDomain .Values.containerLogs.fluentBit.image.repository .Values.containerLogs.fluentBit.image.tag -}}
 {{- end -}}
 
+{{/*
+Helper function to add dualstack endpoints to fluent-bit OUTPUT sections
+*/}}
+{{- define "fluent-bit.add-dualstack-endpoints" -}}
+{{- $config := .config -}}
+{{- if and .Values.useDualstackEndpoint (not (contains "endpoint" $config)) -}}
```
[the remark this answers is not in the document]
the-mann: do we have a test that validates this logic?

duhminick:
It would be worth adding some test scenarios in https://github.com/aws-observability/helm-charts/tree/main/integration-tests/amazon-cloudwatch-observability/terraform/minikube/scenarios for the logic in here imo

charts/amazon-cloudwatch-observability/templates/_helpers.tpl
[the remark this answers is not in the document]
movence: it would be helpful to add before/after FB config with ipv6 enabled similar to how you show agent config in the description

charts/amazon-cloudwatch-observability/templates/windows/fluent-bit-windows-configmap.yaml
```diff
@@ -8,14 +8,14 @@ metadata:
     k8s-app: fluent-bit
 data:
   fluent-bit.conf: |
-    {{- .Values.containerLogs.fluentBit.configWindows.service  | nindent 4 }}
+    {{- include "fluent-bit.add-ipv6-preference" (dict "config" .Values.containerLogs.fluentBit.configWindows.service "useDualstackEndpoint" .Values.useDualstackEndpoint "indent" "  ") | nindent 4 }}
     {{- range $key, $val := .Values.containerLogs.fluentBit.configWindows.extraFiles }}
     @INCLUDE {{ $key }}
     {{- end }}
   parsers.conf: |
```
movence: has this been also tested on an windows instance?

charts/amazon-cloudwatch-observability/templates/linux/fluent-bit-configmap.yaml
```diff
@@ -17,27 +17,27 @@ data:
 {{- if hasPrefix "us-iso-" .Values.region }}
```
movence: not an expert with FB config but what happens if a custom fluentbit config doesn't have `Parsers_File` for some reasons? would it be better to just appned right after `[SERVICE]`?

movence: what do you mean custom config is out of scope? the public doc should have some instructions on how to setup ipv6 with custom configs, right? The new function `fluent-bit.add-ipv6-preference` will skip if a custom config already has dual stack configred.

duhminick (approved):

PR #269, opened 2026-01-23
Update config values

sparrc:
*Description of changes:*

I have made a selection of changes to the default config for the following reasons.

Overall the changes are meant to limit runaway memory usage on the fluent-bit container, by setting mem_buf_limit on all input plugins.

One other thing to consider would be to add two new config parameters: [`workers` for the output plugins and `threaded` for the input plugins](https://docs.fluentbit.io/manual/administration/multithreading), which enables multi-threading on the fluent-bit process to increase efficiency and free up fluent-bit's main thread. These options, however, require that the user is using aws-for-fluent-bit version 3+. Is this an assumption we can safely make?

| Change | Reason |
|--------|--------|
| Flush 5 → 2 | Logs are flushed every second instead of every 5 seconds, reducing size of output plugin buffers |
| storage.max_chunks_up 16 | Limits memory chunks to 16 if filesystem storage is enabled |
| Removed storage.type filesystem from tail inputs | The tail plugin already tracks file offsets via the DB option, so filesystem buffering is redundant and creates excessive disk IO. Filesystem buffering also |
| Mem_Buf_Limit 10MB on tail inputs | These were previously using filesystem storage, so setting a mem_buf_limit to keep memory usage in check |
| Added Mem_Buf_Limit 10MB to systemd inputs | Ensures all inputs have memory limits for consistent behavior |
| retry_limit 5 on all outputs | Retries failed sends up to 5 times before dropping logs |

### DIFF
--- a/charts/amazon-cloudwatch-observability/values.yaml
+++ b/charts/amazon-cloudwatch-observability/values.yaml
@@ -56,15 +56,19 @@ containerLogs:
     config:
       service: |
         [SERVICE]
-          Flush                     5
+          Flush                     2
           Grace                     30
           Log_Level                 error
           Daemon                    off
           Parsers_File              parsers.conf
+          # The following storage options only apply to inputs using storage.type filesystem.
+          # Filesystem storage is not used by default, but these options are specified here
+          # as a recommendation in case any inputs are switched to filesystem storage.
           storage.path              /var/fluent-bit/state/flb-storage/
           storage.sync              normal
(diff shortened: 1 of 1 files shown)


J-Swift:
We are seeing regular OOM kills on our fluent-bit pods and I found this PR while researching after AI basically recommended a similar set of changes. It feels off putting to need to modify this on my end if the underlying isnt configured appropriately

PR #282, opened 2026-03-24
Add OTEL Container Insights infrastructure

louisall:
## Summary
Adds OTEL-based Container Insights support to the helm chart, enabled by default. Deploys node-exporter, kube-state-metrics, and a cluster-scraper (as an `AmazonCloudWatchAgent` CR in the `agents` array) alongside the existing CloudWatch agent to collect Kubernetes metrics via OTLP and export them to the CloudWatch Monitoring endpoint. Includes feature-targeted agent routing where each feature (`containerInsights`, `applicationSignals`, `otelContainerInsights`) declares a `targetAgent` field controlling which agent receives its config.

## Changes

**Chart infrastructure:**
- New OTEL collector configs for per-node scraping (cadvisor, kubeletstats, node-exporter, EFA, EBS CSI, DCGM, Neuron) and cluster-level scraping (kube-state-metrics, API server)
- Node-exporter DaemonSet with TLS, kube-state-metrics Deployment with TLS, cluster-scraper as an `agents` array entry (`mode: deployment`) managed by the operator
- OTEL config merge logic that preserves user-supplied otelConfig alongside generated pipelines (generated config takes precedence on collision), with `fromYaml` error validation
- `containerInsights.enabled` flag to independently control legacy Container Insights pipeline
- Region-aware image resolution via `repositoryDomainMap` (public, China, GovCloud) with `restrictedRepository`/`restrictedTag` for restricted regions
- KSM/node-exporter receivers and pipelines gated by their respective `enabled` flags — disabled components don't produce scrape configs
- `scrapeTimeout` helper using `min(metricResolution, 10s)` to prevent Prometheus rejection when `metricResolution` < 10s
- Fixed `monitorAllServices` logic to use `applicationSignals.enabled` + `targetAgent` routing instead of inspecting `$agent.config` (which is nil for default installs)
- KSM naming aligned with sidecar component pattern (`kube-state-metrics.name`/`kube-state-metrics.serviceAccountName` helpers), RBAC split into separate clusterrole/clusterrolebinding files
- Removed health probes and `health_check` OTEL extensions from agent CRs — operator manages agent health; avoids port 13133 conflict with `hostNetwork: true`
- `otelConfig` field omitted from CRs when no OTEL features target the agent (instead of emitting empty config)

**Integration tests:**
- Updated EKS resource counts and name patterns for new OTEL resources
- Extended minikube default scenario to validate OTEL resources exist
- New `otlp-disabled` and `otlp-custom-otel-config` scenarios
- New feature-targeted scenarios: default, multi-agent, otlp-disabled, split-features, appsignals-disabled, ci-disabled, both-disabled, custom-otel-config, user-config-override
- Added `ValidateDaemonSetExists`, `ValidateServiceExists`, `ValidateServiceAccountExists` helpers to K8sClient

**Documentation:**
- Hierarchical AGENTS.md files at each directory boundary for AI-assisted development context
- `.helmignore` to exclude .md files from chart packaging

## Configuration
New values under `otelContainerInsights`:
- `enabled` (default: true) — gates all OTEL CI resources
- `targetAgent` (default: "cloudwatch-agent") — agent receiving node-level OTEL config
- `clusterScraperAgent` (default: "cloudwatch-agent-cluster-scraper") — agent receiving cluster-level OTEL config
- `metricResolution` (default: "30s") — scrape interval
- `cloudwatchMetricsEndpoint` — optional endpoint override

`kubeStateMetrics` and `nodeExporter` sections at root level with `enabled`, `name`, `image`, `resources`, `serviceAccount`. Resources only created when both the component's `enabled` flag and `otelContainerInsights.enabled` are true.

`containerInsights.enabled` (default: true) independently controls the legacy CI pipeline. Both can run simultaneously.

## Testing
- `make all` passes (helm lint, secret scan, go fmt/lint)
- All 25 minikube scenarios pass (0 failures)
- Deployed and verified on EKS test cluster


louisall:
### Known CI failure: Minikube integration tests timeout

All Minikube scenarios fail with `context deadline exceeded` on the helm install step. This is expected and will resolve once the updated CloudWatch Agent image ships with OTEL component support.

**Root cause:** This PR sets `otelContainerInsights.enabled: true` by default, which deploys a `cloudwatch-agent-cluster-scraper` Deployment. This deployment uses the public CW agent image (`public.ecr.aws/cloudwatch-agent/cloudwatch-agent`), which does not yet include the OTEL components referenced in the generated config (`otlphttp` exporter, `kubeletstats` receiver, `metricstarttime` processor, etc.). The cluster-scraper pod starts, fails config validation, and enters `CrashLoopBackOff`. Since the Terraform helm provider uses `wait=true` by default, it blocks until all helm-managed pods are ready, hits the 300s timeout, and fails.

The Go test assertions themselves pass — they validate resource existence and config content, not pod health. Verified locally by running all 16 scenarios with `--wait=false`: 16/16 pass.

**What will fix it:** When CW agent image released with the required OTEL collector components. Once that image is published to public ECR, the cluster-scraper will start successfully and `wait=true` will no longer timeout.

### Manual validation performed:

- Merge precedence logic verified on EKS cluster with 3 scenarios: OTLP enabled + colliding user config, OTLP disabled + user config, OTLP enabled without user config
- All 16 minikube Go tests pass when deployed with `--wait=false`
- `helm template` dry-runs confirm correct config rendering for all scenarios

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1195,3 +1195,61 @@ neuronMonitor:
...
+  image:
+    repository: kube-state-metrics/kube-state-metrics
+    tag: v2.18.0
+    repositoryDomainMap:
+      public: registry.k8s.io
+      cn-north-1: 918309763551.dkr.ecr.cn-north-1.amazonaws.com.cn
+      cn-northwest-1: 961992271922.dkr.ecr.cn-northwest-1.amazonaws.com.cn
+      us-gov-east-1: 151742754352.dkr.ecr.us-gov-east-1.amazonaws.com
+      us-gov-west-1: 013241004608.dkr.ecr.us-gov-west-1.amazonaws.com
+    restrictedRepository: eks/kube-state-metrics
```
mitali-salvi, line 1203: What does this mean ? same question for ```restrictedTag```

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1195,3 +1195,61 @@ neuronMonitor:
                   - fargate
   nodeSelector:
     kubernetes.io/os: linux
+
+kubeStateMetrics:
+  image:
+    repository: kube-state-metrics/kube-state-metrics
+    tag: v2.18.0
```
mitali-salvi, line 1195: Does the tag need the version ```v``` ?

mitali-salvi: Same comment about node-exporter

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1195,3 +1195,61 @@ neuronMonitor:
...
+      us-gov-west-1: 013241004608.dkr.ecr.us-gov-west-1.amazonaws.com
+    restrictedRepository: eks/prometheus-node-exporter
+    restrictedTag: v1.10.2-eksbuild.8
+  serviceAccount:
+    name:
+
+## Controls the legacy Container Insights pipeline (CloudWatch Agent collecting kubernetes metrics via logs.metrics_collected.kubernetes).
+## When set to false, the kubernetes key is removed from the agent config, disabling legacy CI metrics collection.
+## Can run alongside otelContainerInsights — both pipelines are independent.
+containerInsights:
```
mitali-salvi, line 1268: do we want to remove the container-Inisghts section in the CWA default JSON config ? it may get confusing for customers to enable container insights in 2 places

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1195,3 +1195,61 @@ neuronMonitor:
...
+  enabled: true
+
+## Controls the OTLP-based Container Insights pipeline (node-exporter, kube-state-metrics, cluster-scraper, OTEL collector config).
```
mitali-salvi, line 1279: should these components be enabled by default ? What if the customer has existing workloads deployed already and dont want duplicates ?

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1195,3 +1195,61 @@ neuronMonitor:
...
+## Both containerInsights and otelContainerInsights can be enabled simultaneously — they run as independent pipelines.
+## | containerInsights.enabled | otelContainerInsights.enabled | Result                    |
+## |--------------------------|-------------------------------|---------------------------|
+## | true (default)           | true (default)                | Both pipelines active     |
+## | true                     | false                         | Legacy CI only            |
+## | false                    | true                          | OTLP CI only              |
+## | false                    | false                         | No CI metrics             |
+otelContainerInsights:
+  enabled: true
+  metricResolution: "15s"
```
mitali-salvi: Will this change the metric resolution in all components of the otelContainerInsights pipeline ?

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1195,3 +1195,61 @@ neuronMonitor:
                   - fargate
   nodeSelector:
     kubernetes.io/os: linux
+
+kubeStateMetrics:
```
mitali-salvi, line 1189: How can customers configure the memory/cpu/tolerance for this component ?

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
mitali-salvi, line 1: Whats the reason for naming this file to start with a underscore ?

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
```diff
@@ -0,0 +1,254 @@
+{{- define "otel-container-insights-cluster-scraper.config" -}}
+extensions:
+  sigv4auth/otelci_cwotel:
+    region: {{ .Values.region }}
+    service: monitoring
+  health_check/otelci:
```
mitali-salvi: Is the health check required ?

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
```diff
@@ -97,13 +97,35 @@ spec:
...
+  {{- end }}
   {{- if $agent.config }}
   config: {{ include "cloudwatch-agent.modify-config" (merge (dict "Config" $agent.config) $ ) }}
   {{- else }}
   config: {{ include "cloudwatch-agent.modify-config" (merge (dict "Config" $agent.defaultConfig) $ ) }}
   {{- end }}
-  {{- if $agent.otelConfig }}
-  otelConfig: {{ include "cloudwatch-agent.modify-otel-config" (merge (dict "OtelConfig" $agent.otelConfig) . ) }}
+  {{- if and $.Values.otelContainerInsights.enabled $agent.otelConfig }}
+  otelConfig: {{ include "cloudwatch-agent.modify-otel-config" (merge (dict "OtelConfig" (include "cloudwatch-agent.merge-otel-configs" (dict "Base" (include "otel-container-insights.config" $) "User" $agent.otelConfig))) $) }}
```
mitali-salvi: Should this be conditional as in only merge the otel configs for container insights depending on the mode (daemon-set/deployment)

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
```diff
@@ -0,0 +1,143 @@
...
+    resources: [statefulsets, daemonsets, deployments, replicasets]
+    verbs: [list, watch]
+  - apiGroups: [batch]
+    resources: [cronjobs, jobs]
+    verbs: [list, watch]
+  - apiGroups: [autoscaling]
+    resources: [horizontalpodautoscalers]
+    verbs: [list, watch]
+  - apiGroups: [authentication.k8s.io]
+    resources: [tokenreviews]
```
mitali-salvi: Do we know why is this needed ? especially create permissions

charts/amazon-cloudwatch-observability/templates/linux/otel-container-insights-cluster-scraper-clusterrole.yaml
mitali-salvi, line 1: Are there new permissions here which are not already covered in the existing cluster role ? https://github.com/aws-observability/helm-charts/blob/main/charts/amazon-cloudwatch-observability/templates/cloudwatch-agent-clusterrole.yaml

charts/amazon-cloudwatch-observability/templates/linux/otelci-cluster-scraper-deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "cloudwatch-agent.name" . }}-cluster-scraper
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "amazon-cloudwatch-observability.labels" . | nindent 4 }}
+---
+apiVersion: apps/v1
+kind: Deployment
```
mitali-salvi: Is there a reason this is a separate Deployment workload as opposed to appending to the agent array - https://github.com/aws-observability/helm-charts/blob/main/charts/amazon-cloudwatch-observability/values.yaml#L853

charts/amazon-cloudwatch-observability/templates/cloudwatch-agent-clusterrole.yaml
```diff
@@ -27,6 +27,11 @@ rules:
 - apiGroups: [ "" ]
   resources: [ "nodes/stats", "configmaps", "events" ]
   verbs: [ "create", "get" ]
+{{- if .Values.otelContainerInsights.enabled }}
```
mitali-salvi, line 30: Do we need this along wit the clusterRole here as well ? charts/amazon-cloudwatch-observability/templates/linux/otelci-cluster-scraper-clusterrole.yaml

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -0,0 +1,676 @@
...
+              regex: ${env:K8S_NODE_NAME}
+              action: keep
+            - source_labels: [__meta_kubernetes_pod_container_port_name]
+              regex: metrics
+              action: keep
+
+  kubeletstats/cw_k8s_ci_v0:
+    auth_type: serviceAccount
+    collection_interval: {{ .Values.otelContainerInsights.metricResolution }}
+    endpoint: "https://${env:K8S_NODE_NAME}:10250"
```
JayPolanco: For consistency with cAdvisor, this should also use `HOST_IP` instead of `K8S_NODE_NAME`. If a cluster doesn't have CoreDNS set, scraping will fail as well

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
sky333999: Just a convention for template files.

charts/amazon-cloudwatch-observability/values.yaml
sky333999: The actual ksm and node-exp images tags have `v` in their versions.

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1195,3 +1195,61 @@ neuronMonitor:
...
+      public: registry.k8s.io
+      cn-north-1: 918309763551.dkr.ecr.cn-north-1.amazonaws.com.cn
+      cn-northwest-1: 961992271922.dkr.ecr.cn-northwest-1.amazonaws.com.cn
+      us-gov-east-1: 151742754352.dkr.ecr.us-gov-east-1.amazonaws.com
+      us-gov-west-1: 013241004608.dkr.ecr.us-gov-west-1.amazonaws.com
+    restrictedRepository: eks/kube-state-metrics
+    restrictedTag: v2.18.0-eksbuild.1
+
+nodeExporter:
+  name:
```
sky333999, line 1227: I see `name` and `serviceAccount` exposed here, but not on ksm above. We should be consistent, not just between these two, but ideally across the other existing components as well.

Id recommend adding the common shared set:
* enabled: true
* name:
* image:
* resources: (cpu, mem etc.)
* securityContext
* tolerations (same pattern as others, i.e. commented out by default)
* affinity
* nodeSelector
* service

charts/amazon-cloudwatch-observability/values.yaml
sky333999: Perhaps we need to drop the `agent.defaultConfig` entirely at this point and construct it based on these root level flags (i.e. add a similar boolean for applicationSignals too).

We should discuss the expectation of this flag for customers that have currently overridden the `agent.config` field and passed in their own json config.

charts/amazon-cloudwatch-observability/values.yaml
sky333999: Do we need to expose any settings here besides `enabled`? Any changes should be via the more granular settings above. This is just adding too many layers IMO.

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1195,3 +1195,61 @@ neuronMonitor:
...
+## | true (default)           | true (default)                | Both pipelines active     |
+## | true                     | false                         | Legacy CI only            |
+## | false                    | true                          | OTLP CI only              |
+## | false                    | false                         | No CI metrics             |
+otelContainerInsights:
+  enabled: true
+  metricResolution: "15s"
+  # Override the CloudWatch Metrics OTLP endpoint. If empty, defaults to monitoring.<region>.amazonaws.com.
+  cloudwatchMetricsEndpoint: ""
+  clusterScraper:
```
sky333999: I wouldnt expose these as separate settings as yet another component. `agents` section seems more appropriate to me so it matches all the settings available there.

charts/amazon-cloudwatch-observability/templates/cloudwatch-agent-clusterrole.yaml
sky333999: Yeah this is for the regular daemonset and the other is for the cluster-wide deployment.

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
sky333999: I dont see how this works at this layer. agents can be used to configure multiple CRs today, so technically multiple daemonsets and deployments and now this logic will inject CI into all of them.

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
```diff
@@ -0,0 +1,143 @@
...
+    resources: [certificatesigningrequests]
+    verbs: [list, watch]
+  - apiGroups: [coordination.k8s.io]
+    resources: [leases]
+    verbs: [list, watch]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: cwagent-kube-state-metrics
```
sky333999: All these need to be consistent with existing components. Same name pattern, annotations pattern, labels pattern etc.

As an example, I would have expected this to be `{{ template "amazon-cloudwatch-observability.name" . }}-ksm-role-binding`

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
```diff
@@ -0,0 +1,143 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "amazon-cloudwatch-observability.labels" . | nindent 4 }}
+---
+# ClusterRole is cluster-scoped — prefixed with "cwagent-" to avoid
+# collision with a customer's own kube-state-metrics ClusterRole.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
sky333999: nit: Id keep the pattern consistent. Like role and rolebinding are separate files for node-exporter but combined here for ksm - just match them to what the other components have.

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
```diff
@@ -0,0 +1,143 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cwagent-kube-state-metrics
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "amazon-cloudwatch-observability.labels" . | nindent 4 }}
+    app.kubernetes.io/component: cwagent-kube-state-metrics
+spec:
+  replicas: 1
```
sky333999, line 19: nit: Fine for now, but we should expose replicas, PDB settings and allow for a HPA.

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
```diff
@@ -0,0 +1,143 @@
...
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: cwagent-kube-state-metrics
```
sky333999: Same comment as before - make all of these consistent with existing naming conventions and patterns.

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
```diff
@@ -0,0 +1,143 @@
...
+        kubernetes.io/os: linux
+      {{- dict "component" .Values.kubeStateMetrics "context" . | include "amazon-cloudwatch-observability.common.tolerations" | nindent 6 }}
+      containers:
+        - name: kube-state-metrics
+          image: {{ template "kube-state-metrics.image" . }}
+          ports:
+            - name: http-metrics
+              containerPort: 8080
+              protocol: TCP
+            - name: telemetry
```
sky333999, line 49: Are these self-health prom metrics? Do we need a prom scrape annotation?

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
sky333999: +1. Please make sure the create permissions are absolutely needed.

charts/amazon-cloudwatch-observability/templates/linux/node-exporter-daemonset.yaml
```diff
@@ -0,0 +1,112 @@
...
+              readOnly: true
+              mountPropagation: HostToContainer
+            - name: sys
+              mountPath: /host/sys
+              readOnly: true
+            - name: proc
+              mountPath: /host/proc
+              readOnly: true
+            - name: nodeexportertls
+              mountPath: /etc/node-exporter-cert
```
sky333999, line 69: How does not exporter know to pick this up? Dont see anything in args above.
Is this the default path? If so, can you point me to the docs.

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
```diff
@@ -0,0 +1,143 @@
...
+          resources:
+            limits:
+              cpu: 200m
+              memory: 256Mi
+            requests:
+              cpu: 50m
+              memory: 64Mi
+---
+apiVersion: v1
+kind: Service
```
sky333999, line 89: We have tls for node-exp but not ksm?

charts/amazon-cloudwatch-observability/templates/linux/otelci-cluster-scraper-deployment.yaml
```diff
@@ -0,0 +1,111 @@
+{{- if .Values.otelContainerInsights.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "cloudwatch-agent.name" . }}-cluster-scraper-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "amazon-cloudwatch-observability.labels" . | nindent 4 }}
+data:
+  cwagent-config.json: |
```
sky333999: Do you even need a json config?

charts/amazon-cloudwatch-observability/templates/linux/otelci-cluster-scraper-deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+  replicas: 1
+  strategy:
+    type: Recreate
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ template "cloudwatch-agent.name" . }}-cluster-scraper
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include "otel-container-insights-cluster-scraper.config" . | sha256sum }}
```
sky333999: Do you need to force reconcile it this way? The operator should be doing this - please test it out.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
```diff
@@ -0,0 +1,254 @@
+{{- define "otel-container-insights-cluster-scraper.config" -}}
```
sky333999, line 1: We do the boolean kind check for the one below but not here - please make them consistent.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -0,0 +1,676 @@
+{{- define "otel-container-insights.config" -}}
```
sky333999, line 1: Lets pick one naming convention - otel-container-insights vs otelci and use that everywhere. Like the files use diff conventions. Apply whatever we pick for the field exposed in helm values as well.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
louisall: Done

charts/amazon-cloudwatch-observability/values.yaml
louisall: Addressed in the refactor. agent.defaultConfig has been removed entirely. The CW Agent JSON config is now
dynamically constructed by a build-default-config helper based on the root-level feature flags (
containerInsights.enabled, applicationSignals.enabled). Each flag has a targetAgent field that routes its
config to a specific agent in the agents array. There's no longer a static JSON blob that every agent inherits
— an agent only gets the features that target it. This eliminates the "two places" confusion: the root-level
flags are the single source of truth, and the helper builds the JSON from them.

charts/amazon-cloudwatch-observability/values.yaml
louisall: Enabling otel by default was a Product decision I didnt make. Customers with existing node-exporter or kube-state-metrics deployments can set otelContainerInsights.enabled: false to disable the entire OTEL CI pipeline, or individually set nodeExporter.enabled: false / kubeStateMetrics.enabled: false to skip just those components. We should document this in the chart README as a migration note.

charts/amazon-cloudwatch-observability/values.yaml
louisall: Yes. Weve decided to change it back to 30s though

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
louisall: Addressed by the targetAgent routing. The OTEL config merge is no longer mode-dependent — it's name-
dependent. The build-default-otel-config helper checks which agent name matches otelContainerInsights.targetAgent (node-level config) or otelContainerInsights.clusterScraperAgent (cluster-
level config). An agent that matches neither gets a health-check-only config. So a deployment-mode prometheus-agent won't get OTEL CI config injected — not because of its mode, but because its name doesn't match any targetAgent. This is more flexible than mode-gating since you could have multiple DaemonSets where only one should get OTEL CI.

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
louisall: Removed. The KSM ClusterRole now only has list and watch verbs. The tokenreviews and subjectaccessreviews create permissions were from the upstream kube-state-metrics chart's default RBAC and aren't needed for our use case — we only need KSM to read cluster state, not perform authentication/authorization reviews.

charts/amazon-cloudwatch-observability/templates/linux/otel-container-insights-cluster-scraper-clusterrole.yaml
louisall: Yes — the cluster-scraper ClusterRole has a narrower scope than the main agent ClusterRole. It only needs
get/list/watch on pods, nodes, namespaces, endpoints, services, endpointslices, and get on /metrics. The main agent ClusterRole has additional permissions for things like create on configmaps/events, which the cluster-scraper doesn't need. They're separate because the cluster-scraper uses its own ServiceAccount (
cloudwatch-agent-cluster-scraper) and should follow least-privilege.

charts/amazon-cloudwatch-observability/templates/linux/otelci-cluster-scraper-deployment.yaml
louisall: Addressed — the standalone Deployment has been deleted. The cluster-scraper is now an entry in the agents
array with mode: deployment, rendered as an AmazonCloudWatchAgent CR and managed by the operator like every other agent. This was one of the main goals of the refactor.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
louisall: Done

charts/amazon-cloudwatch-observability/values.yaml
louisall: Addressed — both kubeStateMetrics and nodeExporter now expose the full common set: enabled, name, image, resources, securityContext, tolerations, affinity, nodeSelector, service, serviceAccount. Tolerations follow the existing pattern (commented out by default, inheriting root tolerations).

charts/amazon-cloudwatch-observability/values.yaml
louisall: Kept metricResolution for now — it's the one setting that applies uniformly across all OTEL CI scrape jobs (
both node-level and cluster-level) and doesn't have a natural home in the per-component sections. Moving it to individual component configs would mean duplicating the same value across 9+ scrape configs.
cloudwatchMetricsEndpoint is also kept as it's an endpoint override that applies to all OTEL CI exporters.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
louisall: Standardized on otel-container-insights for file names (matching the tpl files). The values.yaml key remains
otelContainerInsights (camelCase, standard Helm convention for values keys). Internal OTEL component names in the config templates use their own prefixes (cw_k8s_ci_v0 for node-level, otel_container_insights for cluster-level) which are agent-internal identifiers, not user-facing.

charts/amazon-cloudwatch-observability/templates/linux/otelci-cluster-scraper-deployment.yaml
louisall: Removed — the standalone Deployment with its checksum annotation is gone. The cluster-scraper is now an
operator-managed CR, and the operator handles config reconciliation.

charts/amazon-cloudwatch-observability/templates/linux/otelci-cluster-scraper-deployment.yaml
louisall: The standalone ConfigMap was deleted along with the standalone Deployment. The cluster-scraper now gets its config through the CR's config field (dynamically constructed by build-default-config, which gives it
{"agent":{"region":"..."}} since no features target it by name) and otelConfig field (dynamically constructed
by build-default-otel-config, which gives it the cluster-level OTEL CI config).

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
louisall: Fixed — KSM now has TLS. Added --port=8443 and --tls-config=/etc/ksm-web-config/web.yml args, with a
ConfigMap mounting the TLS cert from the shared agent-cert Secret. Readiness probe updated to scheme: HTTPS.

charts/amazon-cloudwatch-observability/values.yaml
louisall: No change for those customers. The CR template checks if $agent.config first — if the user provides a config
override, it's used directly via modify-config (which adds cluster_name and hosted_in enrichments). The
build-default-config helper is only invoked in the else branch. So existing users who set agent.config or per-
agent config see zero behavioral change.

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
```diff
@@ -96,14 +98,35 @@ spec:
   {{- with $agent.affinity }}
   affinity: {{- toYaml . | nindent 4 }}
   {{- end }}
+  {{- if ne ($agent.mode | default "daemonset") "deployment" }}
   hostNetwork: true
+  {{- end }}
+  livenessProbe:
```
mitali-salvi: What's the use-case for adding these probes ? This custom resource is managed by the CWAgent operator so any errors are bubbled up to the helm-chart level

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -913,24 +918,8 @@ agent:
     issuerAnnotations: {}
   serviceAccount:
     name: # override agent service account name
-  config: # optional config that can be provided to override the defaultConfig
-  otelConfig: # optional YAML config that can be provided to combine with the config/defaultConfig
-  defaultConfig:
```
mitali-salvi, line 918: https://github.com/aws-observability/helm-charts/blob/main/charts/amazon-cloudwatch-observability/templates/_helpers.tpl#L37

The helper checks $agent.config | default dict for application_signals keys to decide whether to enable monitorAllServices. 

With this PR, agents that have no explicit config in
values get their config built dynamically by build-default-config at render time — $agent.config remains nil. So $agentConfig is always {}, $hasAppSignals is always false, and
monitorAllServices is forced to false even when applicationSignals.enabled: true. Application Signals auto-monitoring will be silently disabled for all default installations.

Fix: Check applicationSignals.enabled and applicationSignals.targetAgent directly instead of inspecting the agent config dict:
```
{{- range .Values.agents -}}
  {{- $agent := merge . (deepCopy $.Values.agent) -}}
  {{- if and $.Values.applicationSignals.enabled (eq $.Values.applicationSignals.targetAgent $agent.name) -}}
    {{- $hasAppSignals = true -}}
  {{- end -}}
{{- end -}}
```

charts/amazon-cloudwatch-observability/templates/_helpers.tpl
```diff
@@ -489,3 +568,90 @@ Returns auto-generated certificate and CA for admission webhooks.
...
+  {{/* Merge service.pipelines maps */}}
+  {{- if and (hasKey $user.service "pipelines") (hasKey $merged.service "pipelines") -}}
+    {{- $_ := set $merged.service "pipelines" (mustMergeOverwrite $user.service.pipelines $merged.service.pipelines) -}}
+  {{- else if hasKey $user.service "pipelines" -}}
+    {{- $_ := set $merged.service "pipelines" $user.service.pipelines -}}
+  {{- end -}}
+{{- else if hasKey $user "service" -}}
+  {{- $_ := set $merged "service" $user.service -}}
+{{- end -}}
+{{- $merged | toYaml -}}
```
mitali-salvi, line 714: nit -- does this need error validation ?

fromYaml in Helm returns {"Error": "parse error..."} on failure rather than panicking. The helper does not check for this error key after parsing $base or $user. If a user provides
malformed YAML in otelConfig, the merge proceeds with a map containing an Error key, producing a corrupt otelConfig in the CR with no diagnostic message.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
```diff
@@ -0,0 +1,266 @@
...
+            - source_labels: [__meta_kubernetes_service_name]
+              action: keep
+              regex: kubernetes
+            - source_labels: [__meta_kubernetes_endpoint_port_name]
+              action: keep
+              regex: https
+            - target_label: __metrics_path__
+              replacement: /metrics
+
+  prometheus/otel_container_insights_kube_state_metrics:
```
mitali-salvi: this should check if kube state metrics is enabled at the root level in values.yaml

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -0,0 +1,676 @@
...
+{{- end }}
+extensions:
+  sigv4auth/cw_k8s_ci_v0_metrics_dest:
+    region: {{ .Values.region }}
+    service: monitoring
+  health_check/cw_k8s_ci_v0:
+    endpoint: "0.0.0.0:13133"
+
+receivers:
+  prometheus/cw_k8s_ci_v0_node_exporter:
```
mitali-salvi, line 9: same comment as KSM, this is missing a check if node exporter is enabled

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -0,0 +1,676 @@
...
+    config:
+      scrape_configs:
+        - job_name: node-exporter
+          scrape_interval: {{ .Values.otelContainerInsights.metricResolution }}
+          scheme: https
+          tls_config:
+            insecure_skip_verify: true
```
mitali-salvi, line 17: the CA cert is available similar to other dependencies like DCGM, this should use the same cert for TLS

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
```diff
@@ -0,0 +1,266 @@
...
+    config:
+      scrape_configs:
+        - job_name: kube-state-metrics
+          scrape_interval: {{ .Values.otelContainerInsights.metricResolution }}
+          scrape_timeout: 10s
```
mitali-salvi: Edge case might be here where customer sets the metric resolution < 10 seconds, this should default to 10 but should be set to metricResolution if that value is set

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
```diff
@@ -0,0 +1,266 @@
...
+  health_check/otel_container_insights:
+    endpoint: "0.0.0.0:13133"
+
+receivers:
+  prometheus/otel_container_insights_apiserver:
+    config:
+      scrape_configs:
+        - job_name: kubernetes-apiserver
+          scrape_interval: {{ .Values.otelContainerInsights.metricResolution }}
+          scrape_timeout: 10s
```
mitali-salvi: Edge case might be here where customer sets the metric resolution < 10 seconds, this should default to 10 but should be set to metricResolution if that value is set

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
```diff
@@ -65,6 +65,8 @@ data:
 {{- $isROSA := eq $.Values.k8sMode "ROSA" -}}
 {{- range .Values.agents }}
 {{- $agent := merge . (deepCopy $.Values.agent) }}
+{{/* Skip cluster-scraper agent when otelContainerInsights is disabled */}}
+{{- if not (and (not $.Values.otelContainerInsights.enabled) (eq $agent.name $.Values.otelContainerInsights.clusterScraperAgent)) }}
```
mitali-salvi: can be simplified ?
```
{{- if or $.Values.otelContainerInsights.enabled (ne $agent.name $.Values.otelContainerInsights.clusterScraperAgent) }}
```

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1195,3 +1184,101 @@ neuronMonitor:
                   - fargate
   nodeSelector:
     kubernetes.io/os: linux
+
+kubeStateMetrics:
```
mitali-salvi, line 1189: Should this section be under the otelContainerInsights section ? Asking coz customer can disable otelCOntainerInsights at the root level but this will still install the KSM dependency

charts/amazon-cloudwatch-observability/templates/linux/node-exporter-daemonset.yaml
```diff
@@ -0,0 +1,110 @@
+{{- if and .Values.nodeExporter.enabled .Values.otelContainerInsights.enabled }}
+apiVersion: v1
+kind: ServiceAccount
```
mitali-salvi, line 3: nit - for consistency, this should be moved to node-exporter-role or a dedicated node-exporter-rbac similar to KSM

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
louisall: Addressed with target agent routing

charts/amazon-cloudwatch-observability/values.yaml
louisall: Added a comment in values.yaml. These are the alternate image repo and tag used when the deploy region matches a restricted region (China/GovCloud) in repositoryDomainMap. In those regions the upstream public images (e.g. registry.k8s.io/kube-state-metrics/kube-state-metrics) aren't accessible, so the helper swaps to repositoryDomainMap[region]/restrictedRepository:restrictedTag — which points to an EKS-managed mirror in the region's ECR. For all other regions it uses the normal repository/tag from the public domain.

charts/amazon-cloudwatch-observability/values.yaml
louisall: Addressed in the refactor — the cluster-scraper is now an entry in the agents array with mode: deployment.
Customers configure resources, tolerations, affinity, nodeSelector, etc. directly on that agent entry, same as
any other agent CR. This keeps the configuration surface consistent across all agents.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
louisall: Agreed to remove in next rev

charts/amazon-cloudwatch-observability/values.yaml
louisall: Agreed and done — the separate otelContainerInsights.clusterScraper.resources section has been removed. All cluster-scraper pod-level settings are configured through its entry in the agents array, matching the pattern for every other agent CR.

charts/amazon-cloudwatch-observability/values.yaml
louisall: Resources are only created when both kubeStateMetrics.enabled and otelContainerInsights.enabled are true. The root-level placement is intentional for discoverability, same pattern as nodeExporter and dcgmExporter.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
louisall: Added a scrapeTimeout helper that uses min(metricResolution, 10s) — if metricResolution < 10s, timeout matches the interval (prevents Prometheus rejection); otherwise caps at 10s (avoids waiting 30s+ for a single scrape). Applied to both apiserver and KSM scrape configs. The node-level config doesn't set scrape_timeout at all so Prometheus defaults it to scrape_interval automatically — no change needed there.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
louisall: Added a scrapeTimeout helper that uses min(metricResolution, 10s) — if metricResolution < 10s, timeout matches the interval (prevents Prometheus rejection); otherwise caps at 10s (avoids waiting 30s+ for a single scrape). Applied to both apiserver and KSM scrape configs. The node-level config doesn't set scrape_timeout at all so Prometheus defaults it to scrape_interval automatically — no change needed there.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
louisall: Wrapped the KSM receiver, KSM-specific processors, and KSM pipeline in the cluster-scraper OTEL config with {{- if .Values.kubeStateMetrics.enabled }}. When KSM is disabled, the cluster-scraper only runs the apiserver pipeline. Shared processors stay unconditional.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
louisall: Already using the shared amazon-cloudwatch-observability-agent-cert Secret — same one DCGM and neuron-monitor use. No separate cert setup.

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
louisall: Done

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
louisall: Agreed offline. Removed the probes.

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
louisall: Added a TODO comment. Replicas/PDB/HPA for KSM is a follow-up item, not blocking this PR.

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
louisall: Fixed

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
louisall: Removed the prometheus scrape annotations. The cluster-scraper uses static_configs with the KSM service DNS target directly — nothing in the chart does annotation-based discovery. Added a comment on the Service clarifying this. The telemetry port (8081) is KSM's self-health metrics, only exposed on the container for probes, not on the Service — that's intentional.

charts/amazon-cloudwatch-observability/templates/linux/node-exporter-daemonset.yaml
louisall: Both KSM and node-exporter now have their ServiceAccount in the workload file (deployment/daemonset). This keeps all pod-spec-related resources together in one file, with cluster-scoped RBAC (ClusterRole/ClusterRoleBinding) in separate files.

charts/amazon-cloudwatch-observability/templates/linux/node-exporter-daemonset.yaml
louisall: Good catch — it wasn't there when you reviewed. Added --web.config.file=/etc/node-exporter/web.yml to the args now. It's not a default path, node-exporter requires the explicit flag. Docs: https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md

charts/amazon-cloudwatch-observability/templates/_helpers.tpl
louisall: Added error validation after fromYaml calls in merge-otel-configs. Malformed YAML now fails with a descriptive message (e.g. Failed to parse user-supplied otelConfig: ...) instead of silently producing a corrupt merged config.

charts/amazon-cloudwatch-observability/values.yaml
louisall: Fixed. Replaced with a direct check on applicationSignals.enabled + applicationSignals.targetAgent matching. Verified monitorAllServices: true renders correctly with default values.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
louisall: Wrapped KSM receiver/processors/pipeline in the cluster-scraper config with {{- if .Values.kubeStateMetrics.enabled }}, and node-exporter receiver/processor/pipeline in the node-level config with {{- if .Values.nodeExporter.enabled }}. Same pattern already used by dcgm and neuron pipelines.

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
louisall: Fixed. KSM now follows the same sidecar naming pattern as dcgm/neuron/node-exporter — added kube-state-metrics.name and kube-state-metrics.serviceAccountName helpers, replaced all hardcoded cwagent-kube-state-metrics and amazon-cloudwatch-observability-ksm-* references. Default names: kube-state-metrics, kube-state-metrics-service-acct, kube-state-metrics-cluster-role, etc. Also added a kubeStateMetrics.name override in values.yaml for consistency.

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
louisall: Split kube-state-metrics-rbac.yaml into separate kube-state-metrics-clusterrole.yaml and kube-state-metrics-clusterrolebinding.yaml files, moved ServiceAccount into the deployment file. Matches the dcgm/neuron/node-exporter pattern now.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
mitali-salvi: sure but this exported is not using those certs since this config option will skip TLS altogether

mitali-salvi: This should be changed to something like this
```
          tls_config:
            ca_file: /etc/amazon-cloudwatch-observability-agent-client-cert/tls-ca.crt
```

louisall: Done

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
```diff
@@ -0,0 +1,272 @@
+{{- define "otel-container-insights-cluster-scraper.config" -}}
+{{- if not (kindIs "bool" .Values.otelContainerInsights.enabled) }}
+{{- fail "otelContainerInsights.enabled must be a boolean (true/false)" }}
```
mitali-salvi: These templates are only ever called from cloudwatch-agent.build-default-otel-config, which already performs this check before delegating. This can be removed

mitali-salvi: Same comment applies for the node level otel-ci config as well

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
```diff
@@ -0,0 +1,113 @@
...
+            - --port=8443
+            - --tls-config=/etc/ksm-web-config/web.yml
+          ports:
+            - name: https-metrics
+              containerPort: 8443
+              protocol: TCP
+            - name: telemetry
+              containerPort: 8081
+              protocol: TCP
+          readinessProbe:
```
mitali-salvi, line 52:
```suggestion
          readinessProbe:
            httpGet:
              path: /healthz
              port: 8081
              scheme: HTTP
```

mitali-salvi: KSM's telemetry port (8081) is plain HTTP. The TLS config (--tls-config) only applies to the metrics port (8443). Using scheme: HTTPS on an HTTP port causes the probe to fail, making the pod never become ready.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
louisall: Removed

charts/amazon-cloudwatch-observability/templates/linux/kube-state-metrics.yaml
louisall: Moved readinessProbe to port 8443/HTTPS /healthz

sky333999 (approved):

PR #300, opened 2026-04-21
feat(container-insights): Add k8staints processor to all pipelines

miconeilaws:
Add the `k8staints` processor after `k8sattributes/cw_k8s_ci_v0_node` in all OTEL Container Insights pipelines. The processor enriches metrics and logs with node taint data as `k8s.node.taint.<key>` resource attributes.

Renamed from the originally-proposed `k8snodemetadata` to `k8staints` (see [contrib PR #509](https://github.com/amazon-contributing/opentelemetry-collector-contrib/pull/509)) to disambiguate from the existing `nodemetadataenricher` processor (which enriches with EC2 instance / host metadata) and to honestly reflect that this processor only watches and exposes node taints today.

**DaemonSet config:** `local_mode=true` (watches only the local node)
**Cluster-scraper config:** default mode (watches all nodes)

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
+++ b/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
@@ -274,6 +274,8 @@ processors:
       - sources:
           - from: resource_attribute
             name: k8s.node.name
+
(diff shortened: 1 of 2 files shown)


louisall:
3 failing tests unrelated to this change. I did a fix for them anyway. Should be resolved when this is merged: https://github.com/aws-observability/helm-charts/pull/312

PR #310, opened 2026-05-11
Add OTEL-native Container Insights log pipelines (app, host)

louisall:
## Add OTEL log pipelines for Container Insights

Adds application and host log collection via OTEL, running inside the existing CloudWatch Agent DaemonSet. Opt-in via `otelContainerInsights.logs.enabled: true` (requires `otelContainerInsights.enabled: true`).

### What it does

When enabled, the agent collects:
- **Application logs** — all container stdout/stderr from `/var/log/containers/*.log` (excluding agent and system pods)
- **Host logs** — `/var/log/messages`, `/var/log/dmesg`, `/var/log/secure`

Logs are sent to CloudWatch Logs via OTLP with per-node streams:
- `/aws/otel/containerinsights/<cluster>/application`
- `/aws/otel/containerinsights/<cluster>/host`

### Metrics–logs attribute parity

Every log record carries the same resource attributes as the metrics for the same pod/node. This means you can correlate logs and metrics by `k8s.pod.name`, `k8s.namespace.name`, `k8s.workload.name`, `k8s.node.name`, `cloud.resource_id`, pod labels, node labels, etc. — all populated by the same shared `k8sattributes` processor instances (one K8s API watch serves both signals).

### How it works

- `filelog` receiver reads container/host log files
- `container_log_parser` operator strips CRI/Docker envelope (depends on [otel-contrib#533](https://github.com/amazon-contributing/opentelemetry-collector-contrib/pull/533))
- `awscloudwatchlogsprovisioner` extension pre-creates log groups/streams lazily (depends on [agent#2113](https://github.com/aws/amazon-cloudwatch-agent/pull/2113))
- Shared `k8sattributes` processors enrich logs with the same pod/node metadata as metrics
- Transform processors add cluster name, cloud resource ID, workload name — identical to the metrics pipeline

### Config

```yaml
otelContainerInsights:
  enabled: true
  logs:
    enabled: true   # opt-in, default true (no-op when enabled=false)
```

If `enabled` is false, everything under `otelContainerInsights` is ignored (including `logs.enabled`).

### Testing

- `flag_matrix.sh` validates all 8 flag combinations (template-level, wired into `make all`)
- Integration test scenarios: logs-disabled, dual-publish, hybrid-metrics-fluentbit, config-merge-isolation

### Changes since initial review

- `otelContainerInsights.logs` restructured from flat boolean to `otelContainerInsights.logs.enabled` (future-proofing for additional log knobs)
- Removed phantom `dataplane` receivers/pipelines from tests (deferred to future PR)
- Removed `runlogjournal`/`varlogjournal` volume checks (FluentBit volumes, not CW Agent)
- Cleaned up stale cross-flag rule enforcement code and comments — all 8 flag combinations are valid
- Fixed `default_test.go` comment accuracy
- Wired `flag_matrix.sh` into CI via `make helm-flag-matrix` (runs in Build job)
- `filestorage` for checkpoint persistence deferred to follow-up PR (requires writable hostPath — security posture change)
- `queue_size: 500` retained per benchmark data (Appendix C.8 — eliminates drops at X-Large with zero CPU/memory cost)


charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1118,15 +1278,28 @@ applicationSignals:
...
+##   enabled=false              → no OTEL CI (default, v6.x behavior)
+##   enabled=true               → OTEL metrics + OTEL logs
+##   enabled=true, logs=false   → OTEL metrics only
+##
 otelContainerInsights:
-  enabled: true
+  enabled: false
+  ## Enables OTEL-native CI log pipelines (application, host).
+  ## Only takes effect when otelContainerInsights.enabled is true.
+  logs: true
```
sky333999: Can we make this an obj too? So an `enabled` field under this. Keeps it future proof if we need to add more `logs` related knobs.

integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/otlp_logs_disabled_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+// assertLogPipelineAbsent checks that no log-specific OTEL components appear in
+// the generated config. The shared names are gated by otelContainerInsights.logs.
+func assertLogPipelineAbsent(t *testing.T, otelConfig string) {
+	t.Helper()
+	logOnlyFragments := []string{
+		// Receivers
+		"filelog/cw_k8s_ci_v0_app",
+		"filelog/cw_k8s_ci_v0_node",
+		"filelog/cw_k8s_ci_v0_dataplane_containers",
```
sky333999: These `dataplane` ones dont exist in the actual pipeline def I think?

integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/otlp_logs_dual_publish_test.go
```diff
@@ -0,0 +1,159 @@
...
+			break
+		}
+	}
+	if !assert.NotNil(t, ds, "DaemonSet %q not found", dsName) {
+		return
+	}
+
+	requiredVolumes := map[string]bool{
+		"varlog":        false,
+		"runlogjournal": false,
```
sky333999: What is `runlogjournal` volume? Dont see changes for this to the actual volumes on the agent custom resource.

charts/amazon-cloudwatch-observability/tests/flag_matrix.sh
```diff
@@ -0,0 +1,198 @@
...
+        --set "otelContainerInsights.logs=$logs" \
+        --set "containerLogs.enabled=$fb" 2>&1) && exit_code=0 || exit_code=$?
+
+    if [[ "$expected" == "fail" ]]; then
+        if [[ $exit_code -eq 0 ]]; then
+            echo -e "  ${R}FAIL${N}: expected helm template to fail, but it succeeded"
+            fail_count=$((fail_count + 1))
+            return
+        fi
+        if ! grep -q "otelContainerInsights.logs=true requires otelContainerInsights.enabled=true" <<< "$output"; then
```
sky333999: What would print this error message? Dont see it anywhere else in this PR.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -548,9 +696,52 @@ exporters:
...
+    # available as an opt-in bandwidth optimization (~5–10× reduction for typical
+    # container logs) at the cost of +25% agent CPU — customers who raise the CPU
+    # limit accordingly can improve on FluentBit's bandwidth cost. See OTELify CI
+    # Logs Pipeline Optimizations doc for details.
+    compression: none
+    headers:
+      x-aws-log-group: "/aws/otel/containerinsights/{{ .Values.clusterName }}/application"
+      x-aws-log-stream: "${env:K8S_NODE_NAME}-application"
+    sending_queue:
+      queue_size: 500
```
sky333999, line 753: This seems too large for this default? Should we start with `100`?

integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/default_test.go
```diff
@@ -11,6 +11,18 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+// TestDefault validates the out-of-box default install behavior.
+//
+// Default corresponds to state 2 in the CI flag state matrix (see values.yaml):
+//   containerInsights.enabled       = true  (default)
+//   containerLogs.enabled           = true  (default)
+//   otelContainerInsights.enabled   = false (default — OTEL is opt-in)
+//   otelContainerInsights.logs      = false (default — OTEL logs are opt-in)
```
sky333999: This technically should be `true` per our `values.yaml`. Just that its not effective given the flag on the previous line is `false`.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -129,6 +134,52 @@ receivers:
...
+    include:
+      - /var/log/containers/*.log
+    exclude:
+      - /var/log/containers/cloudwatch-agent*
+      - /var/log/containers/fluent-bit*
+      # aws-node and kube-proxy are dataplane components — excluded to match
+      # FluentBit's existing Exclude_Path behavior.
+      - /var/log/containers/aws-node*
+      - /var/log/containers/kube-proxy*
+    start_at: end
```
sky333999, line 176: This actually is key for us to discuss - whats the behavior when the agent pod restarts?

Today fluent-bit keeps track of the offset in storage and can recover gracefully from that point even after a restart. Its the same for CWA's classic logs functionality elsewhere.

So we should try to achieve the same here using something like the `filestorage` extension perhaps - needs some thorough testing.

charts/amazon-cloudwatch-observability/tests/flag_matrix.sh
```diff
@@ -0,0 +1,198 @@
...
+#   containerLogs.enabled
+#
+# This is a template-level test — it does not deploy anything to a real cluster.
+# Minikube integration tests cover deployment; this script fills the gap for:
+#   (a) testing all 8 combinations quickly without minikube
+#   (b) asserting that the two invalid combinations (#3, #4) fail helm template
+#       as expected (minikube terraform harness cannot test "expected failure")
+#
+# Run from the repo root:
+#     bash charts/amazon-cloudwatch-observability/tests/flag_matrix.sh
```
sky333999: Is this meant to be checked in? What going to actually run this in the Github CI? Dont see any changes to our workflows.

If there is value on top of the minikube tests, can we wire it into `helm-lint` or a `flag-matrix-test` make target?

charts/amazon-cloudwatch-observability/values.yaml
louisall: Good shout. Will do

integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/otlp_logs_disabled_test.go
louisall: Removed.

integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/otlp_logs_dual_publish_test.go
louisall: You're right — runlogjournal and varlogjournal are FluentBit volumes (for journald). Removed the journald volume checks from both tests

charts/amazon-cloudwatch-observability/tests/flag_matrix.sh
louisall: Think this just wasn't cleaned up from when we had this rule. miconeil updated to allow all states with these flags (slack feedback about the complicated state rules) but kiro didnt clean it all up. Did a full check for any stale comment references about the old rules.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
louisall: 500 was a deliberate choice from our benchmarking. We ran a queue-size sensitivity sweep at X-Large (50 MB/s): queue_size=50 produced 73 queue-full events (data loss), queue_size=500 eliminated all drops, and queue_size=1000 provided no additional benefit. CPU and memory were flat across all three values. The metrics exporter in this same template uses the OTEL default of 1000 (no explicit config). 500 is already half that and is the minimum that avoids drops under our heaviest tested workload. What do you think — does that rationale hold, or is there a memory concern at lower tiers I'm not seeing?

integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/default_test.go
louisall: good catch, fixed

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
louisall: Yes this was a miss. Right now logs written during an agent restart are lost. As you said we need the otel `filestorage` but theres no way to do it without persisting to file (or somewhere else but file makes most sense). I know ye had said writing to file tends to be an AppSec concern but I don't see anther way around this. In order not to block i'll leave this PR as it is and do the `filestorage` change in a separate follow up PR to decouple.

charts/amazon-cloudwatch-observability/tests/flag_matrix.sh
louisall: Good catch, wired it into CI via a helm-flag-matrix Make target. It now runs as part of make all. No minikube needed — it's pure helm template so it runs in ~10 seconds

louisall:
Failures unrelated. This PR will fix them: https://github.com/aws-observability/helm-charts/pull/312

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -594,9 +745,52 @@ exporters:
     auth:
       authenticator: sigv4auth/cw_k8s_ci_v0_metrics_dest
 
+{{- if .Values.otelContainerInsights.logs.enabled }}
+  otlphttp/cw_k8s_ci_v0_app_logs_dest:
+    endpoint: {{ if .Values.otelContainerInsights.cloudwatchLogsEndpoint }}{{ .Values.otelContainerInsights.cloudwatchLogsEndpoint | quote }}{{ else }}"https://logs.{{ .Values.region }}.amazonaws.com:443"{{ end }}
```
movence: This should handle ADC regions similar how FB does currently

```
{{- if hasKey .Values.adcEndpointOverrides .Values.region }}
    endpoint            logs.${AWS_REGION}.{{ index .Values.adcEndpointOverrides .Values.region }}
{{- end }}
```

charts/amazon-cloudwatch-observability/tests/flag_matrix.sh
movence: this feels redundant with minikube tests getting added. nice to have an extra layer but it is weird we test a feature (flags) at build time. My point is we should be consistent that this is going to be a new pattern we follow by enforcing various rules at build time OR we just stay with the existing pattern of simple minikube tests.

integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/otlp_hybrid_metrics_fluentbit_test.go
```diff
@@ -0,0 +1,114 @@
...
+	assert.True(t, exists, "operator deployment should exist")
+
+	// FluentBit should be deployed.
+	exists, err = k8sClient.ValidateDaemonSetExists(minikube.Namespace, "fluent-bit")
+	assert.NoError(t, err)
+	assert.True(t, exists, "fluent-bit DaemonSet should be deployed when containerLogs.enabled=true")
+
+	// AmazonCloudWatchAgent CR should carry OTEL metrics config but no log pipelines.
+	dynamicClient, err := k8sClient.GetDynamicClient()
+	if !assert.NoError(t, err) {
```
movence: nit: seeing this pattern in other places too. can be simplied with 
```suggestion
	require.NoError(t, err, "failed to get dynamic client")
```

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -586,6 +640,103 @@ processors:
...
+        cloud.region: { enabled: true }
+        cloud.availability_zone: { enabled: true }
+        cloud.account.id: { enabled: true }
+
+  transform/cw_k8s_ci_v0_logs_set_cloud_resource_id:
+    error_mode: ignore
+    log_statements:
+      - context: resource
+        statements:
+          - set(attributes["cloud.resource_id"], Concat(["arn:aws:eks:", attributes["cloud.region"], ":", attributes["cloud.account.id"], ":cluster/", attributes["k8s.cluster.name"]], ""))
```
movence, line 686: non-eks support matters?

spanaik: Don't think we have added support for these yet, we can revisit once we are here. Will add as followup.

charts/amazon-cloudwatch-observability/tests/flag_matrix.sh
spanaik: Lemme take a look if minikube tests cover everything for now and remove this.

spanaik: Agreed — kept it simple by going with consistent minikube-only coverage.

Resolution in [`1d84963`](https://github.com/aws-observability/helm-charts/commit/1d84963):
- Removed `flag_matrix.sh` and the `helm-flag-matrix` Make target
- Added new minikube scenario [`otlp-logs-otel-only`](https://github.com/aws-observability/helm-charts/tree/feature/otelify-ci-logs/integration-tests/amazon-cloudwatch-observability/terraform/minikube/scenarios/otlp-logs-otel-only) for state #7 (`enabled=true, logs=true, containerLogs=false`) — this was the only state #7 was uniquely covered by the matrix script

Minikube coverage of the 8 flag combinations:
| State | enabled / logs / containerLogs | Minikube scenario |
|---|---|---|
| #1 | f / f / f | covered by other "off" scenarios |
| #2 | f / f / t | `default` (default values) |
| #3 | f / t / f | degenerate to #1 (logs ignored when enabled=f) |
| #4 | f / t / t | degenerate to #2 |
| #5 | t / f / f | `otlp-logs-disabled` |
| #6 | t / f / t | `otlp-hybrid-metrics-fluentbit` |
| #7 | t / t / f | `otlp-logs-otel-only` ✨ new |
| #8 | t / t / t | `otlp-logs-dual-publish` |

States #3/#4 are template-pure (silent ignore of `logs.enabled`) and degenerate to existing minikube cases, so no new scenario is needed.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -586,6 +640,103 @@ processors:
...
+
+  batch/cw_k8s_ci_v0_logs_dest:
+    send_batch_size: 500
+    send_batch_max_size: 500
+    timeout: 5s
+{{- end }}
+
 exporters:
   otlphttp/cw_k8s_ci_v0_metrics_dest:
     endpoint: {{ if .Values.otelContainerInsights.cloudwatchMetricsEndpoint }}{{ .Values.otelContainerInsights.cloudwatchMetricsEndpoint | quote }}{{ else }}"https://monitoring.{{ .Values.region }}.amazonaws.com:443"{{ end }}
```
movence, line 732: we should support ADC as well here like logs

movence (approved):

.github/workflows/amazon-cloudwatch-observability-integration-test.yaml
```diff
@@ -68,6 +68,10 @@ jobs:
           - feature-targeted-agent-config-override
           - otlp-custom-otel-config
           - otlp-disabled
+          - otlp-logs-disabled
+          - otlp-hybrid-metrics-fluentbit
+          - otlp-logs-dual-publish
+          - otlp-logs-otel-only
```
sky333999, line 74: Do we need to add `agent-config-merge-isolation` here?

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1118,15 +1118,29 @@ applicationSignals:
...
-## | true                     | false                         | Legacy CI only            |
-## | false                    | true                          | OTLP CI only              |
-## | false                    | false                         | No CI metrics             |
+##
+## ══════════════════════════════════════════════════════════════════════════════
+## OTEL Container Insights
+## ══════════════════════════════════════════════════════════════════════════════
+##
+## When otelContainerInsights.enabled is false (default), everything under this
+## section is ignored. Existing v6.x behavior (ECI metrics + FluentBit logs) is
```
sky333999: nit: Id drop the line about `Existing v6.x behavior (ECI metrics + FluentBit logs) is unchanged`.

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1118,15 +1118,29 @@ applicationSignals:
...
+## unchanged.
+##
+## When enabled, the agent collects metrics via OTEL pipelines. To also collect
+## logs via OTEL, set otelContainerInsights.logs.enabled to true.
```
sky333999: OTel CI logs is enabled by default too, but this makes it sound like it needs to also be opted into.

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1118,15 +1118,29 @@ applicationSignals:
...
+##
+## Supported combinations:
+##   enabled=false              → no OTEL CI (default, v6.x behavior)
+##   enabled=true               → OTEL metrics + OTEL logs
+##   enabled=true, logs=false   → OTEL metrics only
```
sky333999: nit: `enabled=true, logs.enabled=false`

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
```diff
@@ -148,6 +148,15 @@ spec:
   - mountPath: /var/log/pods
     name: criologs
     readOnly: true
+  {{- if and $.Values.otelContainerInsights.enabled $.Values.otelContainerInsights.logs.enabled }}
```
sky333999: Gate this and the one below by also comparing `$agent.name` to `$.Values.otelContainerInsights.targetAgent` in the if condition.

integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/otlp_logs_disabled_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// CWA DaemonSet must not have log-related host mounts.
+	assertNoLogMounts(t, k8sClient, "cloudwatch-agent")
+
+	t.Log("OTLP logs-disabled scenario validation passed")
+}
+
+// assertLogPipelineAbsent checks that no log-specific OTEL components appear in
+// the generated config. The shared names are gated by otelContainerInsights.logs.enabled.
+func assertLogPipelineAbsent(t *testing.T, otelConfig string) {
```
sky333999, line 102: Can we make sure this captures all the fragments exhaustively? Think you might be missing a few.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -586,6 +640,103 @@ processors:
...
+
+  transform/cw_k8s_ci_v0_logs_set_cluster_name:
+    error_mode: ignore
+    log_statements:
+      - context: resource
+        statements:
+          - set(attributes["k8s.cluster.name"], "{{ .Values.clusterName }}")
+          - set(attributes["k8s.node.name"], "${env:K8S_NODE_NAME}")
+
+  resourcedetection/cw_k8s_ci_v0_logs:
```
sky333999: Do we need this duplicated component? Cant we just use `resourcedetection/cw_k8s_ci_v0`for this logs pipeline as well?

integration-tests/amazon-cloudwatch-observability/terraform/minikube/scenarios/agent-config-merge-isolation/values.yaml
```diff
@@ -0,0 +1,14 @@
+region: us-west-2
+clusterName: minikube
+# Test validates OTEL-CI-specific routing (cluster-scraper CR behavior), so
+# explicitly opt into OTEL (off by default per v7.0.0 opt-in design).
```
sky333999: There shouldnt be any mentions of v7.0.0 in the code. In fact any versions for that matter.

charts/amazon-cloudwatch-observability/templates/_helpers.tpl
```diff
@@ -109,9 +109,32 @@ Logic:
...
+    - otelContainerInsights.enabled         — OTEL Container Insights (metrics)
+    - otelContainerInsights.logs.enabled    — OTEL log pipelines
+
+  All flag combinations are valid. Notable behaviors:
+    - otelCI.logs.enabled=true without otelCI.enabled=true is a no-op
+      (logs config is only rendered when the parent OTEL CI pipeline is active)
+    - otelCI.enabled=true + containerLogs.enabled=true = dual-publish
+      (both OTEL and FluentBit log pipelines run simultaneously)
+*/ -}}
+{{- if not (kindIs "bool" $ctx.Values.containerInsights.enabled) }}
```
sky333999: Having containerInsights.enabled and containerLogs.enabled validated inside a `build-default-otel-config` helper seems odd.

Can we extract all of these into a new helper towards the top, like a `chart.validate-flags` perhaps.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
sky333999: Got it. Think thats reasonable for now.

charts/amazon-cloudwatch-observability/tests/flag_matrix.sh
sky333999: Can we update PR desc to reflect the same?

.github/workflows/amazon-cloudwatch-observability-integration-test.yaml
louisall: Yes, added

charts/amazon-cloudwatch-observability/values.yaml
louisall: Done

charts/amazon-cloudwatch-observability/values.yaml
louisall: Good catch, reworded to clarify logs are on by default when enabled=true, and the opt-out path is logs.enabled=false.

charts/amazon-cloudwatch-observability/values.yaml
louisall: fixed

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
louisall: Done, both the volumeMount and volume are now gated by (`eq $agent.name $.Values.otelContainerInsights.targetAgent`). Cluster-scraper won't get the /var/log mount

integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/otlp_logs_disabled_test.go
louisall: Added all missing log-specific processors and extensions.  List is now exhaustive against the template.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
louisall: You're right, config was identical. Removed the duplicate and both log pipelines now reference the shared

integration-tests/amazon-cloudwatch-observability/terraform/minikube/scenarios/agent-config-merge-isolation/values.yaml
louisall: Removed all version references (v7.0.0, v6.x) from scenario values and chart comments.

charts/amazon-cloudwatch-observability/templates/_helpers.tpl
louisall: Done, extracted into `cloudwatch-agent.validate-flags` helper, called from build-default-otel-config. Validation logic is now separate from config generation.

integration-tests/amazon-cloudwatch-observability/validations/minikube/scenarios/otlp_logs_disabled_test.go
```diff
@@ -0,0 +1,172 @@
...
+//   - Log-only sigv4auth extension is absent; metrics-side sigv4auth is present
+//   - CWA DaemonSet does not mount /var/log or journald host paths
+//   - FluentBit DaemonSet is not rendered
+func TestOTLPLogsDisabled(t *testing.T) {
+	k8sClient, err := util.NewK8sClient()
+	require.NoError(t, err, "failed to create k8s client")
+
+	// Namespace + operator sanity.
+	ns, err := k8sClient.GetNamespace(minikube.Namespace)
+	assert.NoError(t, err)
```
mitali-salvi: Might wanna reconsider updating this to ```require.NoError(t, err)``` coz line 38 will panic if ns == nil

mitali-salvi: Same applies for other files including ```validations/minikube/scenarios/otlp_logs_disabled_test.go, otlp_logs_dual_publish_test.go, otlp_logs_otel_only_test.go, otlp_hybrid_metrics_fluentbit_test.go```

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -151,6 +159,52 @@ receivers:
...
+    include:
+      - /var/log/containers/*.log
+    exclude:
+      - /var/log/containers/cloudwatch-agent*
+      - /var/log/containers/fluent-bit*
+      # aws-node and kube-proxy are dataplane components — excluded to match
+      # FluentBit's existing Exclude_Path behavior.
+      - /var/log/containers/aws-node*
+      - /var/log/containers/kube-proxy*
+    start_at: end
```
mitali-salvi, line 176: nit - I know its referenced in the PR description but might be worth adding a comment here as well
```No checkpoint persistence — logs during agent downtime are lost. Filestorage support is tracked in follow-up PR.```

File: charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl [hunk 24]
```diff
@@ -586,6 +640,89 @@ processors:
...
+          - set(attributes["k8s.workload.type"], "Job") where attributes["k8s.job.name"] != nil and attributes["k8s.workload.type"] == nil
+          - set(attributes["k8s.workload.name"], attributes["k8s.cronjob.name"]) where attributes["k8s.workload.name"] == nil and attributes["k8s.cronjob.name"] != nil
+          - set(attributes["k8s.workload.type"], "CronJob") where attributes["k8s.cronjob.name"] != nil and attributes["k8s.workload.type"] == nil
+          - set(attributes["k8s.workload.name"], attributes["k8s.replicaset.name"]) where attributes["k8s.workload.name"] == nil and attributes["k8s.replicaset.name"] != nil
+          - set(attributes["k8s.workload.type"], "ReplicaSet") where attributes["k8s.replicaset.name"] != nil and attributes["k8s.workload.type"] == nil
+          # Derive service.name from k8s.workload.name (OTEL logs semconv).
+          # Logs need service.name; metrics use k8s.workload.name directly.
+          - set(attributes["service.name"], attributes["k8s.workload.name"]) where attributes["service.name"] == nil and attributes["k8s.workload.name"] != nil
+
+  transform/cw_k8s_ci_v0_logs_set_cluster_name:
```
mitali-salvi: nit - this seems like its setting the node name attribute as well, might wanna rename this

sky333999 (approved):

PR #324, opened 2026-06-30
Per-node allocation, SM/PM CRD bundling, and otel-CI scraping

wenegiemepraise:
*Issue #, if available:*

*Description of changes:*
Chart support for the new Target Allocator (TA) capabilities in amazon-cloudwatch-agent-operator: per-node allocation, zero-step ServiceMonitor/PodMonitor CRD bundling

### DIFF
--- a/charts/amazon-cloudwatch-observability/values.yaml
+++ b/charts/amazon-cloudwatch-observability/values.yaml
@@ -765,12 +765,11 @@ agent:
     config:
     targetAllocator:
       enabled: false
-      # allocationStrategy determines how the Target Allocator distributes scrape
-      # targets across CloudWatch Agent pods. "per-node" assigns each target to the
-      # agent on the same node (avoids cross-node/cross-AZ scrape traffic); targets
-      # without a resolvable node fall back to consistent-hashing. "consistent-hashing"
-      # spreads targets across all agents by hash.
-      allocationStrategy: per-node
+      # For agents that enable the Target Allocator directly (not via
+      # otelContainerInsights, which sources it from prometheusScrape.allocationStrategy).
+      # consistent-hashing = hash targets across all agents; per-node = assign to the
+      # agent on the same node.
+      allocationStrategy: consistent-hashing
       image:
         repository: cloudwatch-agent-target-allocator
         tag: 1.0.0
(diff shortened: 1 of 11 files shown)


charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -786,6 +765,12 @@ agent:
...
       enabled: false
+      # allocationStrategy determines how the Target Allocator distributes scrape
+      # targets across CloudWatch Agent pods. "per-node" assigns each target to the
+      # agent on the same node (avoids cross-node/cross-AZ scrape traffic); targets
+      # without a resolvable node fall back to consistent-hashing. "consistent-hashing"
+      # spreads targets across all agents by hash.
+      allocationStrategy: per-node
       image:
         repository: cloudwatch-agent-target-allocator
         tag: 1.0.0
```
Aakash-Dantre, line 775: how are we planning to release the cloudwatch-agent-target-allocator per node strategy? this wont work with 1.0.0 if I am correct right?

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -786,6 +765,12 @@ agent:
     config:
     targetAllocator:
       enabled: false
```
Aakash-Dantre, line 767: targetAllocator.enabled: false is silently ignored when otelContainerInsights.enabled: true ? 

it can be counter intuitive IMO.

What will customer need to setup if they want to have otelContainerInsights.enabled but dont have the scraping enabled?

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1171,3 +1167,20 @@ otelContainerInsights:
...
+## installed by prometheus-operator), so we never clobber/downgrade it. They are
+## annotated helm.sh/resource-policy: keep so `helm uninstall` does NOT delete
+## them (which would cascade-delete every ServiceMonitor/PodMonitor in the
+## cluster).
+prometheusCRDs:
+  ## When to bundle the SM/PM CRDs:
+  ##   auto   - bundle only when otelContainerInsights.enabled is true (default)
+  ##   always - always bundle (still skipped if an unmanaged copy exists)
+  ##   never  - never bundle
+  install: auto
```
Aakash-Dantre: ## Suggestion: simplify the SM/PM scraping flags in `values.yaml`

Enabling this today spans several flags whose interactions aren't obvious — otelContainerInsights.enabled, serviceMonitor.enabled, podMonitor.enabled, prometheusCRDs.install, and the underlying
   targetAllocator.enabled / prometheusCR.enabled. E.g. serviceMonitor.enabled: true is inert unless CI is on, prometheusCRDs: auto only bundles when CI is on, and targetAllocator.enabled: false is silently
   overridden by CI. That's a lot of coupled surface for what's really one decision.

   Could we make the common case a single switch: otelContainerInsights.enabled: true ⇒ SM/PM scraping works end-to-end (CRDs bundled + Target Allocator + prometheusCR + RBAC), false ⇒ none of it renders — no other
   flags needed. Then expose one override to opt out of just the Prometheus-CR scraping while keeping the rest of CI:

PR #329, opened 2026-07-09
[amazon-cloudwatch-observability] feat: add ServiceMonitor/PodMonitor v2 OTLP scraping path

wenegiemepraise:
## Summary
Add the ServiceMonitor/PodMonitor **v2 OTLP scraping path** to the chart: a `prometheuscr`
receiver + pipeline driven by the Target Allocator's PrometheusCR discovery, plus the
enablement flags and gating helper. This is the foundation that lets the CloudWatch Agent
scrape community `monitoring.coreos.com` ServiceMonitor/PodMonitor CRs into CloudWatch via
the OTEL Container Insights path.

## What this enables
When `otelContainerInsights.enabled` is set, the chart wires the Target Allocator +
`prometheuscr` receiver so SM/PM CRs are discovered and scraped (default consistent-hashing
allocation). ServiceMonitor and PodMonitor discovery can each be toggled via
`otelContainerInsights.serviceMonitor.enabled` / `podMonitor.enabled` (default on).

## Scope
- `prometheuscr` receiver + metrics pipeline in the OTEL Container Insights config
- `otelCIScrape` gating helper
- `otelContainerInsights.serviceMonitor/podMonitor.enabled` values (only take effect when
  `otelContainerInsights.enabled`)
- Read-only clusterrole additions for the discovery path

This is the **base of a small stack**; per-node allocation + node-enrichment transforms land
in a follow-up that builds on this path.

## Testing
`helm template`:
- `otelContainerInsights.enabled=true` → the `prometheuscr` receiver/pipeline renders.
- default (disabled) → the path does not render (0 references).

`helm lint` clean (pre-existing `fluent-bit-windows` region warning is unrelated).

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/_helpers.tpl
+++ b/charts/amazon-cloudwatch-observability/templates/_helpers.tpl
@@ -151,6 +151,20 @@ Logic:
 {{- end -}}
 
 {{/*
+Returns "true" when otelContainerInsights-driven ServiceMonitor/PodMonitor scraping
+applies to the given agent. True when otelContainerInsights is enabled, the agent is
+the configured targetAgent, and at least one of serviceMonitor/podMonitor is enabled.
+Accepts a dict with "agentName" (string) and "context" (root context $).
+*/}}
+{{- define "cloudwatch-agent.otelCIScrapeEnabled" -}}
+{{- $ctx := .context -}}
+{{- $agentName := .agentName -}}
+{{- if and $ctx.Values.otelContainerInsights.enabled (eq $agentName $ctx.Values.otelContainerInsights.targetAgent) (or $ctx.Values.otelContainerInsights.serviceMonitor.enabled $ctx.Values.otelContainerInsights.podMonitor.enabled) -}}
+true
+{{- end -}}
+{{- end -}}
+
+{{/*
 Helper function to modify cloudwatch-agent config
 */}}
 {{- define "cloudwatch-agent.config-modifier" -}}
--- a/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
+++ b/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
@@ -43,6 +43,23 @@ receivers:
             - targets:
                 - ${env:HOST_IP}:10250
 
+{{- if or .Values.otelContainerInsights.serviceMonitor.enabled .Values.otelContainerInsights.podMonitor.enabled }}
+  # ServiceMonitor/PodMonitor scraping via the Target Allocator (prometheusCR discovery).
+  # The Target Allocator deployed for the targetAgent serves the scrape jobs derived from
+  # ServiceMonitor/PodMonitor CRs; this receiver pulls this collector's assigned shard and
+  # routes the series into the v2 OTLP pipeline (-> CloudWatch/Zeus). Requires the Target
+  # Allocator + prometheusCR to be enabled for the targetAgent and the POD_NAME env (set below).
+  prometheus/cw_k8s_ci_v0_prometheuscr:
+    target_allocator:
+      endpoint: https://{{ .Values.otelContainerInsights.targetAgent }}-target-allocator-service:80
+      interval: {{ .Values.otelContainerInsights.metricResolution }}
+      collector_id: ${env:POD_NAME}
+      tls:
+        ca_file: /etc/amazon-cloudwatch-observability-agent-cert/tls-ca.crt
+        cert_file: /etc/amazon-cloudwatch-observability-agent-ta-client-cert/client.crt
+        key_file: /etc/amazon-cloudwatch-observability-agent-ta-client-cert/client.key
+{{- end }}
+
   {{- if .Values.dcgmExporter.enabled }}
   prometheus/cw_k8s_ci_v0_dcgm:
     config:
@@ -328,6 +345,16 @@ processors:
           - set(attributes["cloudwatch.source"], "cloudwatch-agent")
           - set(attributes["cloudwatch.solution"], "k8s-otel-container-insights")
           - set(attributes["cloudwatch.pipeline"], "efa")
+
+  transform/cw_k8s_ci_v0_set_scope_prometheuscr:
+    error_mode: ignore
+    metric_statements:
+      - context: scope
+        statements:
+          - set(scope.schema_url, "")
+          - set(attributes["cloudwatch.source"], "cloudwatch-agent")
+          - set(attributes["cloudwatch.solution"], "k8s-otel-container-insights")
+          - set(attributes["cloudwatch.pipeline"], "prometheus-cr")
 
   transform/cw_k8s_ci_v0_set_scope_ebs_csi:
     error_mode: ignore
@@ -828,6 +855,23 @@ service:
         - batch/cw_k8s_ci_v0_metrics_dest
       exporters:
         - otlphttp/cw_k8s_ci_v0_metrics_dest
+
+{{- if or .Values.otelContainerInsights.serviceMonitor.enabled .Values.otelContainerInsights.podMonitor.enabled }}
+    metrics/cw_k8s_ci_v0_prometheuscr:
+      receivers: [prometheus/cw_k8s_ci_v0_prometheuscr]
+      # Phase 1: minimal chain to get ServiceMonitor/PodMonitor series flowing to v2.
+      # Richer OTel label/metadata enrichment (k8sattributes pod/node, workload, etc.)
+      # is intentionally deferred to the Phase 3 enrichment work.
+      processors:
+        - filter/cw_k8s_ci_v0_scrape_metadata
+        - metricstarttime/cw_k8s_ci_v0
+        - transform/cw_k8s_ci_v0_set_cluster_name
+        - transform/cw_k8s_ci_v0_set_scope_prometheuscr
+        - resourcedetection/cw_k8s_ci_v0
+        - batch/cw_k8s_ci_v0_metrics_dest
+      exporters:
+        - otlphttp/cw_k8s_ci_v0_metrics_dest
+{{- end }}
 
     {{- if .Values.dcgmExporter.enabled }}
     metrics/cw_k8s_ci_v0_dcgm:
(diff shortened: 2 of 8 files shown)


charts/amazon-cloudwatch-observability/templates/operator-deployment.yaml
```diff
@@ -37,7 +37,7 @@ spec:
         - "--auto-instrumentation-dotnet-image={{ template "auto-instrumentation-dotnet.image" . }}"
         - "--auto-instrumentation-nodejs-image={{ template "auto-instrumentation-nodejs.image" . }}"
         - "--target-allocator-image={{ template "target-allocator.image" (merge .Values.agent.prometheus.targetAllocator.image (dict "region" $.Values.region)) }}"
-        - "--feature-gates=operator.autoinstrumentation.multiinstrumentation,operator.autoinstrumentation.multiinstrumentation.skipcontainervalidation"
+        - "--feature-gates=operator.autoinstrumentation.multi-instrumentation,operator.autoinstrumentation.multi-instrumentation.skip-container-validation"
```
musa-asad: These `--feature-gates` values are hyphenated (`multi-instrumentation`), but the pinned operator only registers the unhyphenated forms, so it rejects the gate and `CrashLoopBackOff`s and nothing reconciles. This also looks out of scope here. Can we restore the unhyphenated line?

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1161,6 +1161,17 @@ otelContainerInsights:
   ## Only takes effect when otelContainerInsights.enabled is true.
   logs:
     enabled: true
+  ## Prometheus ServiceMonitor scraping (monitoring.coreos.com/v1) via the Target
+  ## Allocator's prometheusCR discovery. Defaults to true: when
+  ## otelContainerInsights is enabled, ServiceMonitor scraping is on unless
+  ## explicitly disabled here. Only takes effect when otelContainerInsights.enabled is true.
+  serviceMonitor:
```
musa-asad, line 1168: `serviceMonitor.enabled` and `podMonitor.enabled` default to true, so the moment someone turns on `otelContainerInsights` they silently get `prometheusCR` discovery, whose CRD and RBAC this PR doesn't set up yet. Could we default them to false, or gate the path on the CRDs being present?

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -329,6 +346,16 @@ processors:
           - set(attributes["cloudwatch.solution"], "k8s-otel-container-insights")
           - set(attributes["cloudwatch.pipeline"], "efa")
 
+  transform/cw_k8s_ci_v0_set_scope_prometheuscr:
```
musa-asad, line 350: The scraping receiver and pipeline are guarded on `serviceMonitor.enabled`/`podMonitor.enabled`, but the `transform/cw_k8s_ci_v0_set_scope_prometheuscr` processor isn't, so turning both monitors off leaves it orphaned and the collector warns. Can we guard it the same way via `otelCIScrapeEnabled`?

charts/amazon-cloudwatch-observability/templates/target-allocator-clusterrole.yaml
```diff
@@ -1,6 +1,7 @@
 {{- if .Values.agent.enabled }}
 {{- range $i, $customAgent := .Values.agents }}
-{{- if and (and (hasKey ($customAgent.prometheus) "targetAllocator") (hasKey ($customAgent.prometheus.targetAllocator) "enabled")) $customAgent.prometheus.targetAllocator.enabled }}
+{{- $otelCIScrape := eq (include "cloudwatch-agent.otelCIScrapeEnabled" (dict "agentName" $customAgent.name "context" $)) "true" }}
```
musa-asad, line 11: This renders the ClusterRole inside `range .Values.agents` with a static name, so two matching agents produce duplicate ClusterRoles and Kubernetes just keeps the last write. Could we namespace the name per agent (same for the ClusterRoleBinding), or document that only one agent is supported?

charts/amazon-cloudwatch-observability/templates/_helpers.tpl
```diff
@@ -191,6 +191,20 @@ Logic:
...
+{{/*
+Returns "true" when otelContainerInsights-driven ServiceMonitor/PodMonitor scraping
+applies to the given agent. True when otelContainerInsights is enabled, the agent is
+the configured targetAgent, and at least one of serviceMonitor/podMonitor is enabled.
+Accepts a dict with "agentName" (string) and "context" (root context $).
+*/}}
+{{- define "cloudwatch-agent.otelCIScrapeEnabled" -}}
+{{- $ctx := .context -}}
+{{- $agentName := .agentName -}}
+{{- if and $ctx.Values.otelContainerInsights.enabled (eq $agentName $ctx.Values.otelContainerInsights.targetAgent) (or $ctx.Values.otelContainerInsights.serviceMonitor.enabled $ctx.Values.otelContainerInsights.podMonitor.enabled) -}}
```
musa-asad, line 203: This helper reads `otelContainerInsights.serviceMonitor.enabled` directly. Fine today since the PR ships those keys as defaults, but it goes nil if a user replaces the whole map, so a `dig` with a default fallback might be safer.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -43,6 +43,23 @@ receivers:
...
+{{- if or .Values.otelContainerInsights.serviceMonitor.enabled .Values.otelContainerInsights.podMonitor.enabled }}
+  # ServiceMonitor/PodMonitor scraping via the Target Allocator (prometheusCR discovery).
+  # The Target Allocator deployed for the targetAgent serves the scrape jobs derived from
+  # ServiceMonitor/PodMonitor CRs; this receiver pulls this collector's assigned shard and
+  # routes the series into the v2 OTLP pipeline (-> CloudWatch/Zeus). Requires the Target
+  # Allocator + prometheusCR to be enabled for the targetAgent and the POD_NAME env (set below).
+  prometheus/cw_k8s_ci_v0_prometheuscr:
+    target_allocator:
+      endpoint: https://{{ .Values.otelContainerInsights.targetAgent }}-target-allocator-service:80
+      interval: {{ .Values.otelContainerInsights.metricResolution }}
```
musa-asad, line 55: `interval` here is the collector to Target Allocator refresh cadence, but it reuses `otelContainerInsights.metricResolution`, which is really the scrape cadence (`prometheusCR.scrapeInterval`). Worth a comment or a separate field so the two can diverge later?

charts/amazon-cloudwatch-observability/values.yaml
wenegiemepraise: The default-on behavior is intentional, the product goal is that enabling otelContainerInsights gives you working SM/PM scraping out of the box, and the surrounding work is built to make that safe: the operator's CRD-resilience change(#394) makes missing CRDs non-fatal (no crash/loop), and CRD bundling (#331) puts the CRDs in-cluster. So I'd like to keep SM/PM default-on rather than invert to opt-in.

charts/amazon-cloudwatch-observability/templates/_helpers.tpl
wenegiemepraise: Since the subkeys ship as chart defaults and Helm deep-merges them, they're present under any partial override, so this is safe as-is I'll leave the direct access

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
wenegiemepraise: Agreed they're distinct knobs. I'd rather not add a second field preemptively, but the shared value being non-obvious is fair

wenegiemepraise:
Makes sense. The pinned cloudwatch-agent-target-allocator:1.0.0 predates the discovery path, and enabling scraping against it CrashLoops the TA rather than degrading, so a discovery-capable TA image must be pinned before this path is enabled (landing in the same release that exposes it, or the path stays off until then).

musa-asad (dismissed):

movence (commented):
Every other Container Insights metrics pipeline runs `clear_schema_url`, `set_cloud_resource_id`, and `awsattributelimit` before `batch`, but this prometheuscr one skips all three. That means ServiceMonitor/PodMonitor metrics ship uncapped and without the schema-url and cloud-resource-id normalization the rest of the pipelines get. Could we add those same three processors here?

(`_otel-container-insights-config.tpl`, prometheuscr pipeline ~L863)

movence (commented):
This loops over the raw `.Values.agents` entries, but the CR template merges the shared `agent` defaults first (`mergeOverwrite (deepCopy .Values.agent)` at `cloudwatch-agent-custom-resource.yaml:67`). So if someone turns on the target allocator or prometheusCR through the shared `agent` block, the CR gets created but this RBAC does not, and the TA's CRD informer ends up denied at runtime. Can we merge `.Values.agent` here the same way?

(`target-allocator-clusterrole.yaml` ~L10)

wenegiemepraise:
Fixed the issue around: `clear_schema_url, set_cloud_resource_id, and awsattributelimit` in the latest commit

For the issue with the looping over the raw `.Values.agents` I dealt with it in PR no.339: https://github.com/aws-observability/helm-charts/pull/339

charts/amazon-cloudwatch-observability/templates/target-allocator-clusterrole.yaml
```diff
@@ -1,6 +1,22 @@
 {{- if .Values.agent.enabled }}
+{{- /*
+  All Target Allocators (targetAgent and clusterScraperAgent) default to the same
+  ServiceAccount (target-allocator-service-acct; see operator serviceaccount.go), so a single
+  ClusterRole covers every TA. Pre-scan the agents to decide whether any TA (and any
+  prometheusCR discovery) is enabled, then render exactly once to avoid duplicate-name objects.
+*/}}
+{{- $needsTA := false }}
+{{- $needsCR := false }}
 {{- range $i, $customAgent := .Values.agents }}
```
movence, line 10: This loops over the raw `.Values.agents` entries, but the CR template merges the shared `agent` defaults first (`mergeOverwrite (deepCopy .Values.agent)` at `cloudwatch-agent-custom-resource.yaml:67`). So if someone turns on the target allocator or prometheusCR through the shared `agent` block, the CR gets created but this RBAC does not, and the TA's CRD informer ends up denied at runtime. Can we merge `.Values.agent` here the same way?

Aakash-Dantre:

## PR Checklist

PR #330, opened 2026-07-09
[amazon-cloudwatch-observability] feat: support per-node Target Allocator allocation

wenegiemepraise:
## Summary
Add **per-node** Target Allocator allocation support to the chart: default the allocation
strategy to `per-node` on the OTEL Container Insights scraping path, and add node-enrichment
transforms (`set_node_name`/`promote_node_name`) that stamp the scraping agent's own node as
`k8s.node.name`. Together with the `target_node` relabel on the SM/PM path, this lets per-node
scraping be verified end-to-end (`target_node == k8s.node.name`) and avoids cross-node/cross-AZ
scrape traffic.

> **Depends on #329** (SM/PM v2 OTLP scraping path) — the node transforms attach to the
> `prometheuscr` pipeline introduced there. Until #329 merges, this PR's diff includes it.
>
> **Companion to operator PR** aws/amazon-cloudwatch-agent-operator (per-node allocation
> strategy) — the chart default `per-node` requires the per-node strategy registered in that
> operator/TA build. Consistent-hashing remains the fallback for node-less targets.

## Scope
- `allocationStrategy` defaults to `per-node` on the CI scraping path (overridable)
- `set_node_name` / `promote_node_name` transforms on the `prometheuscr` metrics pipeline
- CRD allocation-strategy enum widened to include `per-node`

## Testing
`helm template` with `otelContainerInsights.enabled=true`:
- renders `allocationStrategy: "per-node"`
- renders the `set_node_name` / `promote_node_name` transforms on the prometheuscr pipeline

`helm lint` clean (pre-existing `fluent-bit-windows` region warning is unrelated).

### DIFF
--- a/charts/amazon-cloudwatch-observability/crds/cloudwatch.aws.amazon.com_amazoncloudwatchagents.yaml
+++ b/charts/amazon-cloudwatch-observability/crds/cloudwatch.aws.amazon.com_amazoncloudwatchagents.yaml
@@ -6405,9 +6405,10 @@ spec:
                   allocationStrategy:
                     description: |-
                       AllocationStrategy determines which strategy the target allocator should use for allocation.
-                      The current option is consistent-hashing.
+                      The options are consistent-hashing and per-node.
                     enum:
                     - consistent-hashing
+                    - per-node
                     type: string
                   enabled:
                     description: Enabled indicates whether to use a target allocation
--- a/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
+++ b/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
@@ -43,6 +43,23 @@ receivers:
             - targets:
                 - ${env:HOST_IP}:10250
 
+{{- if or .Values.otelContainerInsights.serviceMonitor.enabled .Values.otelContainerInsights.podMonitor.enabled }}
+  # ServiceMonitor/PodMonitor scraping via the Target Allocator (prometheusCR discovery).
+  # The Target Allocator deployed for the targetAgent serves the scrape jobs derived from
+  # ServiceMonitor/PodMonitor CRs; this receiver pulls this collector's assigned shard and
+  # routes the series into the v2 OTLP pipeline (-> CloudWatch/Zeus). Requires the Target
+  # Allocator + prometheusCR to be enabled for the targetAgent and the POD_NAME env (set below).
+  prometheus/cw_k8s_ci_v0_prometheuscr:
+    target_allocator:
+      endpoint: https://{{ .Values.otelContainerInsights.targetAgent }}-target-allocator-service:80
+      interval: {{ .Values.otelContainerInsights.metricResolution }}
+      collector_id: ${env:POD_NAME}
+      tls:
+        ca_file: /etc/amazon-cloudwatch-observability-agent-cert/tls-ca.crt
+        cert_file: /etc/amazon-cloudwatch-observability-agent-ta-client-cert/client.crt
+        key_file: /etc/amazon-cloudwatch-observability-agent-ta-client-cert/client.key
+{{- end }}
+
   {{- if .Values.dcgmExporter.enabled }}
   prometheus/cw_k8s_ci_v0_dcgm:
     config:
@@ -328,6 +345,16 @@ processors:
           - set(attributes["cloudwatch.source"], "cloudwatch-agent")
           - set(attributes["cloudwatch.solution"], "k8s-otel-container-insights")
           - set(attributes["cloudwatch.pipeline"], "efa")
+
+  transform/cw_k8s_ci_v0_set_scope_prometheuscr:
+    error_mode: ignore
+    metric_statements:
+      - context: scope
+        statements:
+          - set(scope.schema_url, "")
+          - set(attributes["cloudwatch.source"], "cloudwatch-agent")
+          - set(attributes["cloudwatch.solution"], "k8s-otel-container-insights")
+          - set(attributes["cloudwatch.pipeline"], "prometheus-cr")
 
   transform/cw_k8s_ci_v0_set_scope_ebs_csi:
     error_mode: ignore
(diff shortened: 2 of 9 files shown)


charts/amazon-cloudwatch-observability/templates/operator-deployment.yaml
```diff
@@ -37,7 +37,7 @@ spec:
         - "--auto-instrumentation-dotnet-image={{ template "auto-instrumentation-dotnet.image" . }}"
         - "--auto-instrumentation-nodejs-image={{ template "auto-instrumentation-nodejs.image" . }}"
         - "--target-allocator-image={{ template "target-allocator.image" (merge .Values.agent.prometheus.targetAllocator.image (dict "region" $.Values.region)) }}"
-        - "--feature-gates=operator.autoinstrumentation.multiinstrumentation,operator.autoinstrumentation.multiinstrumentation.skipcontainervalidation"
+        - "--feature-gates=operator.autoinstrumentation.multi-instrumentation,operator.autoinstrumentation.multi-instrumentation.skip-container-validation"
```
musa-asad: Same `--feature-gates` issue as #329: these are the hyphenated `multi-instrumentation` IDs, but the pinned operator only knows the unhyphenated forms and `CrashLoopBackOff`s on them. Since it came in from #329, could we fix it there and rebase?

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -786,6 +786,12 @@ agent:
     config:
     targetAllocator:
       enabled: false
+      # allocationStrategy determines how the Target Allocator distributes scrape
+      # targets across CloudWatch Agent pods. "per-node" assigns each target to the
+      # agent on the same node (avoids cross-node/cross-AZ scrape traffic); targets
+      # without a resolvable node fall back to consistent-hashing. "consistent-hashing"
+      # spreads targets across all agents by hash.
+      allocationStrategy: per-node
```
musa-asad: Defaulting `allocationStrategy` to `per-node` turns it on for every Target Allocator, and it's a silent no op that doesn't fail safe under image skew (a newer operator against an older allocator just `CrashLoopBackOff`s). Could we keep `consistent-hashing` as the default and make `per-node` opt in until the images support it?

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
```diff
@@ -110,21 +110,46 @@ spec:
...
   {{- end }}
-  {{- if $agent.prometheus.targetAllocator.enabled }}
+  {{- if $taEnabled }}
   targetAllocator:
-    enabled: {{ $agent.prometheus.targetAllocator.enabled | default false }}
+    enabled: true
     image: {{ template "target-allocator.image" (merge $agent.prometheus.targetAllocator.image (dict "region" $.Values.region)) }}
-    allocationStrategy: "consistent-hashing"
-    {{- if $agent.prometheus.targetAllocator.prometheusCR.enabled }}
+    allocationStrategy: {{ $agent.prometheus.targetAllocator.allocationStrategy | default "per-node" | quote }}
```
musa-asad: This `allocationStrategy` emit is under `$taEnabled` but not `$otelCIScrape`, so it also flips the static `scrape_configs` path to `per-node`, which is beyond this PR's scope. Could we gate `per-node` on `$otelCIScrape` and leave the rest on `consistent-hashing`? The `| default` looks redundant too since `values.yaml` already sets it.

charts/amazon-cloudwatch-observability/crds/cloudwatch.aws.amazon.com_amazoncloudwatchagents.yaml
```diff
@@ -6405,9 +6405,10 @@ spec:
                   allocationStrategy:
                     description: |-
                       AllocationStrategy determines which strategy the target allocator should use for allocation.
-                      The current option is consistent-hashing.
+                      The options are consistent-hashing and per-node.
                     enum:
                     - consistent-hashing
+                    - per-node
```
musa-asad, line 6411: The `per-node` enum only lands in `crds/`, which Helm applies on install but not on `helm upgrade`, so upgrading an existing release leaves the old CRD in place and it rejects `per-node`. Could we move the CRD into `templates/` with `helm.sh/resource-policy: keep`?

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -829,6 +856,27 @@ service:
       exporters:
         - otlphttp/cw_k8s_ci_v0_metrics_dest
 
+{{- if or .Values.otelContainerInsights.serviceMonitor.enabled .Values.otelContainerInsights.podMonitor.enabled }}
+    metrics/cw_k8s_ci_v0_prometheuscr:
+      receivers: [prometheus/cw_k8s_ci_v0_prometheuscr]
+      # Phase 1 base chain + node enrichment: set_node_name/promote_node_name stamp
+      # resource.attributes["k8s.node.name"] = the scraping agent's own node
+      # (${env:K8S_NODE_NAME}). Combined with the target_node label from the SM/PM
```
musa-asad: This note says `per-node` is verified end to end via `target_node` matching `k8s.node.name`, but the chart stamps `k8s.node.name` and never actually produces `target_node` (that has to come from the user's SM/PM `relabelings`). Could we reword it as a prerequisite the user supplies rather than something the chart outputs?

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
wenegiemepraise: allocationStrategy now defaults to per-node only on the $otelCIScrape path and consistent-hashing otherwise

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
wenegiemepraise: Rather than just reword it, I removed the node-stamping from the SM/PM path entirely — since per-node placement is already proven via the Target Allocator logs, the set_node_name/promote_node_name stamp on the prometheusCR pipeline wasn't earning its keep. As you noted, it unconditionally overwrote k8s.node.name with the scraping agent's node, which is only correct when per-node placement succeeds (and mislabels/clobbers under fallback). The shared transform definitions stay for the node-level pipelines (cadvisor, node_exporter, etc.) where the agent's node genuinely is the metric's node. That also drops the misleading "verified end-to-end" note.

charts/amazon-cloudwatch-observability/values.yaml
wenegiemepraise: Per-node is intentionally the default, and I'd like to keep it that way rather than ship the feature opt-in. Per-node allocation is the whole point of this change: each DaemonSet agent scrapes only the targets on its own node, which eliminates the cross-node/cross-AZ scrape traffic that consistent-hashing incurs. Defaulting to consistent-hashing would mean the headline behavior is off unless a user knows to flip it, the opposite of "works out of the box."
But if this is a major blocker I could change

charts/amazon-cloudwatch-observability/crds/cloudwatch.aws.amazon.com_amazoncloudwatchagents.yaml
wenegiemepraise: Yeah, moving it to templates/ works great for brand-new CRDs — that's actually why the ServiceMonitor/PodMonitor ones already live there. The problem is AmazonCloudWatchAgent isn't new; it's shipped via crds/ for a long time, so it's already on every existing install without Helm's ownership labels. The moment we move that same CRD into templates/, helm upgrade sees an object it doesn't "own" and hard-fails with an invalid-ownership-metadata error — which is worse than the silent field-pruning we're trying to fix (and resource-policy: keep only affects uninstall, so it doesn't help). So for this one I'm going to document the manual kubectl apply on upgrade (Helm's own recommendation for crds/) and track the proper fix — a pre-upgrade hook that applies the CRD — as a separate change we can actually test against a live upgrade.

musa-asad (approved):

Aakash-Dantre:

## PR Checklist

PR #327, opened 2026-07-06
feat(observability): grant Target Allocator CRD watch RBAC

wenegiemepraise:
## Summary
  Grant the Target Allocator (TA) `get;list;watch` on `customresourcedefinitions`
  (`apiextensions.k8s.io`), scoped under the existing prometheus-CR RBAC gate

  > **Companion to operator PR <https://github.com/aws/amazon-cloudwatch-agent-operator/pull/394>** ("target-allocator: tolerate missing
  > ServiceMonitor/PodMonitor CRDs"). That change makes the TA watch the SM/PM
  > CustomResourceDefinitions so it can start/stop each monitor informer as the CRD
  > appears or disappears. This PR supplies the RBAC that watch requires.
  
  ## Why
  The operator-side change adds an informer over `CustomResourceDefinition` objects.
  Without a grant on `customresourcedefinitions`, that informer is denied at runtime:
  
  Failed to watch *v1.CustomResourceDefinition: customresourcedefinitions.apiextensions.k8s.io is forbidden:
  User "system:serviceaccount:amazon-cloudwatch:cloudwatch-agent-target-allocator" cannot list resource "customresourcedefinitions"
  
  The grant is read-only and cluster-scoped (CRDs are cluster-scoped), the minimum
  required for the watch.
  
  ## Scope / gating
  Added inside the existing `prometheusCR.enabled` gate, right beside the SM/PM
  (`monitoring.coreos.com`) rule — so it renders only when the prometheus-CR scraping
  path is enabled, and is inert otherwise. No dependency on the CRD-bundling or flag
  work; this slots into the RBAC gate already present on `main`.
  
  ## Testing
  `helm template` on the TA ClusterRole:
  - **prometheusCR enabled** → clusterrole includes both the `monitoring.coreos.com`
    (podmonitors/servicemonitors) rule **and** the new `apiextensions.k8s.io`
    (customresourcedefinitions) rule.
  - **prometheusCR disabled** (TA still enabled, control) → clusterrole renders with
    **neither** rule (`customresourcedefinitions: 0`, `podmonitors: 0`) — correctly gated.

PR #331, opened 2026-07-09
[amazon-cloudwatch-observability] feat: bundle ServiceMonitor/PodMonitor CRDs (zero-step install)

wenegiemepraise:
## Summary
Bundle the community `monitoring.coreos.com` **ServiceMonitor/PodMonitor CRDs** in the chart
as templated, gated resources — so customers who don't already run prometheus-operator get
them with **no manual CRD prerequisite** (matching the prometheus-operator experience). This
completes the "zero-step install" of the SM/PM scraping path.

## Contents
- `templates/prometheus-crds/{servicemonitor,podmonitor}-crd.yaml` — the CRD schemas
  (vendored/pinned), **templated + gated**, and **skipped when an unmanaged copy already
  exists** (Helm `lookup`) so a prometheus-operator-managed CRD is never clobbered. Annotated
  `helm.sh/resource-policy: keep` so uninstall does not cascade-delete the SM/PM CRs.
- `values.yaml` — `prometheusCRDs.install` (`auto|always|never`, default `auto`); `auto`
  bundles only when `otelContainerInsights` is enabled.
- `_helpers.tpl` — `prometheusCRDsEnabled` gating helper.
- `tests/prometheus_crds_matrix.sh` — render tests for the gating.

The Target Allocator CRD-watch **RBAC is intentionally not here** — it ships in #327.

## Dependencies
Part of the SM/PM scraping stack; depends on **#329** (v2 OTLP scraping path) and **#330**
(per-node). Until those merge, this PR's diff includes them. Reviewable bottom-up.

## Testing
`helm template`:
- `otelContainerInsights.enabled=true` → **2** CRD manifests render.
- default (disabled) → **0** CRD manifests render (correctly gated).

(The skip-if-exists `lookup` is a no-op under `helm template`/`--dry-run` by design; it takes
effect on real `install`/`upgrade`.)

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/_helpers.tpl
+++ b/charts/amazon-cloudwatch-observability/templates/_helpers.tpl
@@ -151,6 +151,20 @@ Logic:
 {{- end -}}
 
 {{/*
+Returns "true" when otelContainerInsights-driven ServiceMonitor/PodMonitor scraping
+applies to the given agent. True when otelContainerInsights is enabled, the agent is
+the configured targetAgent, and at least one of serviceMonitor/podMonitor is enabled.
+Accepts a dict with "agentName" (string) and "context" (root context $).
+*/}}
+{{- define "cloudwatch-agent.otelCIScrapeEnabled" -}}
+{{- $ctx := .context -}}
+{{- $agentName := .agentName -}}
+{{- if and $ctx.Values.otelContainerInsights.enabled (eq $agentName $ctx.Values.otelContainerInsights.targetAgent) (or $ctx.Values.otelContainerInsights.serviceMonitor.enabled $ctx.Values.otelContainerInsights.podMonitor.enabled) -}}
+true
+{{- end -}}
+{{- end -}}
+
+{{/*
 Helper function to modify cloudwatch-agent config
 */}}
 {{- define "cloudwatch-agent.config-modifier" -}}
--- a/charts/amazon-cloudwatch-observability/templates/_helpers.tpl
+++ b/charts/amazon-cloudwatch-observability/templates/_helpers.tpl
@@ -5,6 +5,25 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{/*
+Whether to bundle the community ServiceMonitor/PodMonitor CRDs. Honours
+.Values.prometheusCRDs.install: "always" => true; "never" => empty;
+"auto" (default) => true only when otelContainerInsights.enabled is true.
+Returns the string "true" when CRDs should be rendered, empty otherwise.
+*/}}
+{{- define "amazon-cloudwatch-observability.prometheusCRDsEnabled" -}}
+{{- $install := "auto" -}}
+{{- if hasKey .Values "prometheusCRDs" -}}
+{{- $install = (.Values.prometheusCRDs.install | default "auto") -}}
+{{- end -}}
+{{- if eq $install "always" -}}
+true
+{{- else if eq $install "never" -}}
+{{- else if .Values.otelContainerInsights.enabled -}}
+true
+{{- end -}}
+{{- end -}}
+
 {{- define "amazon-cloudwatch-observability.common.tolerations" -}}
 {{- $tolerations := .context.Values.tolerations }}
 {{- if .component }}
--- a/charts/amazon-cloudwatch-observability/templates/_helpers.tpl
+++ b/charts/amazon-cloudwatch-observability/templates/_helpers.tpl
@@ -19,8 +19,12 @@ Returns the string "true" when CRDs should be rendered, empty otherwise.
 {{- if eq $install "always" -}}
 true
 {{- else if eq $install "never" -}}
-{{- else if .Values.otelContainerInsights.enabled -}}
+{{- else if eq $install "auto" -}}
+{{- if .Values.otelContainerInsights.enabled -}}
 true
+{{- end -}}
+{{- else -}}
+{{- fail (printf "prometheusCRDs.install must be one of \"auto\", \"always\", or \"never\", got: %s" $install) -}}
 {{- end -}}
 {{- end -}}
 
--- a/charts/amazon-cloudwatch-observability/templates/operator-deployment.yaml
+++ b/charts/amazon-cloudwatch-observability/templates/operator-deployment.yaml
@@ -37,7 +37,7 @@ spec:
         - "--auto-instrumentation-dotnet-image={{ template "auto-instrumentation-dotnet.image" . }}"
         - "--auto-instrumentation-nodejs-image={{ template "auto-instrumentation-nodejs.image" . }}"
         - "--target-allocator-image={{ template "target-allocator.image" (merge .Values.agent.prometheus.targetAllocator.image (dict "region" $.Values.region)) }}"
-        - "--feature-gates=operator.autoinstrumentation.multi-instrumentation,operator.autoinstrumentation.multi-instrumentation.skip-container-validation"
+        - "--feature-gates=operator.autoinstrumentation.multiinstrumentation,operator.autoinstrumentation.multiinstrumentation.skipcontainervalidation"
         command:
         - /manager
         name: manager
--- a/charts/amazon-cloudwatch-observability/templates/operator-deployment.yaml
+++ b/charts/amazon-cloudwatch-observability/templates/operator-deployment.yaml
@@ -37,7 +37,7 @@ spec:
         - "--auto-instrumentation-dotnet-image={{ template "auto-instrumentation-dotnet.image" . }}"
         - "--auto-instrumentation-nodejs-image={{ template "auto-instrumentation-nodejs.image" . }}"
         - "--target-allocator-image={{ template "target-allocator.image" (merge .Values.agent.prometheus.targetAllocator.image (dict "region" $.Values.region)) }}"
-        - "--feature-gates=operator.autoinstrumentation.multiinstrumentation,operator.autoinstrumentation.multiinstrumentation.skipcontainervalidation"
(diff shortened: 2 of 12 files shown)


charts/amazon-cloudwatch-observability/templates/operator-deployment.yaml
```diff
@@ -37,7 +37,7 @@ spec:
         - "--auto-instrumentation-dotnet-image={{ template "auto-instrumentation-dotnet.image" . }}"
         - "--auto-instrumentation-nodejs-image={{ template "auto-instrumentation-nodejs.image" . }}"
         - "--target-allocator-image={{ template "target-allocator.image" (merge .Values.agent.prometheus.targetAllocator.image (dict "region" $.Values.region)) }}"
-        - "--feature-gates=operator.autoinstrumentation.multiinstrumentation,operator.autoinstrumentation.multiinstrumentation.skipcontainervalidation"
+        - "--feature-gates=operator.autoinstrumentation.multi-instrumentation,operator.autoinstrumentation.multi-instrumentation.skip-container-validation"
```
musa-asad: Same `--feature-gates` issue as #329: hyphenated `multi-instrumentation` IDs the pinned operator doesn't register, so it `CrashLoopBackOff`s. Since it comes from #329, could we fix it there and rebase?

charts/amazon-cloudwatch-observability/templates/prometheus-crds/servicemonitor-crd.yaml
```diff
@@ -0,0 +1,1442 @@
+{{- /*
+ServiceMonitor CRD (monitoring.coreos.com/v1), pinned to prometheus-operator v0.91.0 to
```
musa-asad: The header pins these CRDs to prometheus operator `v0.91.0`, but go.mod is actually on `v0.92.0`, so the API server prunes any field added between the two from a user's CR. Could we refetch both CRDs from the `v0.92.0` tag (or fix the header)? Same in `podmonitor-crd.yaml`.

charts/amazon-cloudwatch-observability/templates/prometheus-crds/servicemonitor-crd.yaml
```diff
@@ -0,0 +1,1442 @@
...
+installed by prometheus-operator is never clobbered while our own copy is still
+upgraded on `helm upgrade`. helm.sh/resource-policy: keep prevents `helm uninstall`
+from cascade-deleting every ServiceMonitor in the cluster.
+
+To update: re-fetch the pinned upstream CRD and re-apply the metadata edits
+(helm.sh/resource-policy annotation + chart labels).
+*/ -}}
+{{- if eq (include "amazon-cloudwatch-observability.prometheusCRDsEnabled" .) "true" }}
+{{- $crdName := "servicemonitors.monitoring.coreos.com" }}
+{{- $existing := lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" $crdName }}
```
musa-asad, line 17: `lookup` is empty under `helm template`, so the bundled CRD always applies and can downgrade a GitOps user's newer schema, and the adoption check only reads `meta.helm.sh/release-name`. Could we add a clean opt out and also match `meta.helm.sh/release-namespace` so a cluster scoped CRD can't get cross adopted?

charts/amazon-cloudwatch-observability/tests/prometheus_crds_matrix.sh
```diff
@@ -0,0 +1,78 @@
...
+# always: bundle even with otelCI off.
+expect_count "always + otelCI off => ServiceMonitor CRD bundled" "$SM_CRD" 1 --set otelContainerInsights.enabled=false --set prometheusCRDs.install=always
+expect_count "always + otelCI off => PodMonitor CRD bundled"     "$PM_CRD" 1 --set otelContainerInsights.enabled=false --set prometheusCRDs.install=always
+
+printf "\n${Y}== resource-policy keep on bundled CRDs ==${N}\n"
+expect_count "bundled CRDs carry resource-policy keep" 'helm.sh/resource-policy: keep' 2 --set otelContainerInsights.enabled=true
+
+printf "\n${Y}== Target Allocator CRD RBAC ==${N}\n"
+# The TA needs customresourcedefinitions get;list;watch on the prometheus-CR path.
+expect_count "CRD RBAC present when otelCI on"  "$CRD_RBAC" 1 --set otelContainerInsights.enabled=true
```
musa-asad, line 74: The `CRD RBAC` assertion on line 73 expects `customresourcedefinitions`, but that RBAC lands in a separate PR so this branch renders zero, and `set -euo pipefail` fails the whole suite right here. Could we move those two assertions into the RBAC PR, or guard them until the rule exists?

charts/amazon-cloudwatch-observability/tests/prometheus_crds_matrix.sh
```diff
@@ -0,0 +1,78 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
musa-asad, line 2: This new script uses bash only features but has no shebang, and the `SM_CRD`/`PM_CRD` grep patterns are unanchored with unescaped dots that can over count. Could we add `#!/usr/bin/env bash` and anchor the patterns with `^`/`$` and escaped dots?

charts/amazon-cloudwatch-observability/templates/_helpers.tpl
```diff
@@ -5,6 +5,25 @@ Expand the name of the chart.
...
 {{- end }}
 
+{{/*
+Whether to bundle the community ServiceMonitor/PodMonitor CRDs. Honours
+.Values.prometheusCRDs.install: "always" => true; "never" => empty;
+"auto" (default) => true only when otelContainerInsights.enabled is true.
+Returns the string "true" when CRDs should be rendered, empty otherwise.
+*/}}
+{{- define "amazon-cloudwatch-observability.prometheusCRDsEnabled" -}}
+{{- $install := "auto" -}}
```
musa-asad, line 15: `prometheusCRDsEnabled` only matches `always` and `never` exactly, so anything else falls through and bundles the CRDs, meaning a typo like `nevr` silently bundles when otelCI is on. Could we add a `fail` branch for anything that isn't `auto`, `always`, or `never`, like the sibling key already does?

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1171,3 +1188,20 @@ otelContainerInsights:
...
+## (monitoring.coreos.com/v1). The Target Allocator consumes these CRs; bundling
+## the CRDs removes the manual prerequisite of installing them yourself.
+##
+## The CRDs are shipped as templated manifests (so Helm upgrades them) and are
+## skipped if a copy already exists that this release does not manage (e.g. one
+## installed by prometheus-operator), so we never clobber/downgrade it. They are
+## annotated helm.sh/resource-policy: keep so `helm uninstall` does NOT delete
+## them (which would cascade-delete every ServiceMonitor/PodMonitor in the
+## cluster).
+prometheusCRDs:
```
musa-asad, line 1205: This adds `prometheusCRDs.install` at the top level, but a later PR moves it under `otelContainerInsights.prometheusScrape.crds.install` with no alias, so someone who sets `never` now has it silently ignored after upgrading. Could we hold off calling it stable, or add an alias in that later PR?

charts/amazon-cloudwatch-observability/templates/target-allocator-clusterrole.yaml
```diff
@@ -22,7 +23,7 @@ rules:
     verbs: ["get", "list", "watch"]
   - nonResourceURLs: ["/metrics"]
     verbs: ["get"]
-  {{- if and (hasKey ($customAgent.prometheus.targetAllocator) "prometheusCR") $customAgent.prometheus.targetAllocator.prometheusCR.enabled }}
+  {{- if or (and (hasKey ($customAgent.prometheus) "targetAllocator") (hasKey ($customAgent.prometheus.targetAllocator) "prometheusCR") $customAgent.prometheus.targetAllocator.prometheusCR.enabled) $otelCIScrape }}
```
musa-asad: This branch's TA `ClusterRole` has the SM/PM rule but not a `customresourcedefinitions` rule, which lives in #327, so on its own the `target-allocator` can't `list` CRDs to start discovery. Worth making sure #327 lands alongside this so they gate together.

charts/amazon-cloudwatch-observability/tests/prometheus_crds_matrix.sh
wenegiemepraise: The CRD_RBAC assertions are correct, they only fail because this branch hasn't re-synced onto #329 yet, where the CRD watch RBAC now lives

charts/amazon-cloudwatch-observability/templates/operator-deployment.yaml
wenegiemepraise: Will do

charts/amazon-cloudwatch-observability/templates/_helpers.tpl
wenegiemepraise: split auto into its own branch and added a catch-all else that fails with prometheusCRDs.install must be one of "auto", "always", or "never", got: <value>, matching the metricResolution sibling.

charts/amazon-cloudwatch-observability/values.yaml
wenegiemepraise: I'll ship an alias when #332 relocates it: otelContainerInsights.prometheusScrape.crds.install will fall back to the legacy prometheusCRDs.install and emit a deprecation warning, so a user's never keeps working across the upgrade

charts/amazon-cloudwatch-observability/templates/target-allocator-clusterrole.yaml
wenegiemepraise: handled on #329

musa-asad (approved):

Aakash-Dantre:

## PR Checklist

PR #333, opened 2026-07-09
Add AKS as a supported k8sMode

movence:
## Summary
Adds AKS to k8sMode (EKS | ROSA | K8S | AKS). AKS has no EC2 instance role, so the agent and fluent-bit authenticate to AWS via workload-identity federation:

- Project an SA token (aud sts.amazonaws.com) at /var/run/secrets/aws/token via the aws-iam-token volume
- Set AWS_WEB_IDENTITY_TOKEN_FILE + AWS_ROLE_ARN for the SDK's AssumeRoleWithWebIdentity
- Set RUN_IN_AKS=True for agent platform detection
- New roleArn value, required when k8sMode: AKS
- Wired into both the AmazonCloudWatchAgent CR and the fluent-bit DaemonSet, gated behind $isAKS (mirrors $isROSA).

## Testing
- helm lint clean; EKS/ROSA/K8S renders semantically identical to main
- roleArn guard fails the render when AKS mode has no roleArn
- AKS (live, k8s 1.35, OIDC + workload identity): agent + fluent-bit 1/1 Running, zero auth errors; CI metrics and container logs landing in CloudWatch
- EKS (live, test-mc): upgraded with no regression — agent + fluent-bit unaffected (no AKS wiring), all signals (CI metrics, App Signals, X-Ray, logs) still flowing

Requires the paired agent change that skips the oidctoken path on AKS https://github.com/aws/amazon-cloudwatch-agent/pull/2183. Operator prereq: IAM OIDC provider + role trust-scoped to system:serviceaccount:<ns>:cloudwatch-agent / aud=sts.amazonaws.com, passed as roleArn.

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/linux/fluent-bit-daemonset.yaml
+++ b/charts/amazon-cloudwatch-observability/templates/linux/fluent-bit-daemonset.yaml
@@ -1,6 +1,7 @@
 {{- if .Values.containerLogs.enabled }}
 {{- $clusterName := .Values.clusterName | required ".Values.clusterName is required." -}}
 {{- $region := .Values.region | required ".Values.region is required." -}}
+{{- $isAKS := eq .Values.k8sMode "AKS" -}}
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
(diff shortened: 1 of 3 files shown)


charts/amazon-cloudwatch-observability/templates/linux/fluent-bit-daemonset.yaml
```diff
@@ -1,6 +1,7 @@
 {{- if .Values.containerLogs.enabled }}
 {{- $clusterName := .Values.clusterName | required ".Values.clusterName is required." -}}
 {{- $region := .Values.region | required ".Values.region is required." -}}
+{{- $isAKS := eq .Values.k8sMode "AKS" -}}
```
mitali-salvi, line 4: Missing validation for missing Role ARN similar to the CWA template

```
{{- $isAKS := eq $.Values.k8sMode "AKS" -}}
{{- if $isAKS }}{{- $_ := $.Values.roleArn | required ".Values.roleArn is required when k8sMode is AKS." -}}{{- end }}
```
Helm should render the CWA template before the fluent-bit one but helps to keep validations consistent across YAMLs

mitali-salvi (approved):

jefchien (approved):

PR #332, opened 2026-07-09
[amazon-cloudwatch-observability] refactor: single-switch SM/PM scraping under otelContainerInsights.prometheusScrape

wenegiemepraise:
## Summary
Collapse the SM/PM scraping enablement into a **single switch**. Previously, turning on the
feature spanned several interacting flags (`prometheusCRDs.install`,
`otelContainerInsights.serviceMonitor/podMonitor.enabled`, deep
`agent.prometheus.targetAllocator.*`). This consolidates them under one nested bundle:

```yaml
otelContainerInsights:
  enabled: true                 # the one switch: CRDs + Target Allocator + prometheusCR + RBAC
  prometheusScrape:
    enabled: true               # single opt-out: false => OTEL CI without SM/PM scraping
    allocationStrategy: per-node
    crds:
      install: auto             # auto|always|never
    serviceMonitor: { enabled: true }
    podMonitor:     { enabled: true }
```

`otelContainerInsights.enabled: true` now makes SM/PM scraping work end to end; a single
`prometheusScrape.enabled: false` opts out while keeping the rest of OTEL Container Insights.

## Notes for reviewers
- **API/UX change.** These keys never shipped in a released chart version, so there's no
  external contract to preserve — the old `prometheusCRDs` / `otelContainerInsights.
  serviceMonitor|podMonitor` top-level keys are moved, not aliased.
- Gating helpers in `_helpers.tpl` (`otelCIScrapeEnabled`, `prometheusCRDsEnabled`) now read
  from the `prometheusScrape` bundle.

## Dependencies
Top of the helm stack — refactors keys introduced by **#329** (v2 OTLP path), **#330**
(per-node) and **#331** (CRD bundling). Merge those first; this PR's diff includes them until
they land.

## Testing
`helm template`:
- `otelContainerInsights.enabled=true` → CRDs render (2), prometheuscr path renders,
  `allocationStrategy: "per-node"`.
- `+ prometheusScrape.enabled=false` → prometheuscr and bundled CRDs both drop to 0
  (scraping off, rest of CI intact).

`helm lint` clean (pre-existing `fluent-bit-windows` region warning is unrelated).

PR #339, opened 2026-07-15
[amazon-cloudwatch-observability] Route SM/PM to cluster-scraper by annotation

wenegiemepraise:
## Summary
Let a ServiceMonitor/PodMonitor be routed to the central **cluster-scraper** agent (instead of the default per-node agent) via the `cloudwatch.aws/scraper: cluster-scraper` **annotation** on the monitor CR.

## What
- Give the `clusterScraperAgent` its own Target Allocator + prometheusCR (`otelCIScrapeEnabled` now covers it) with `scraperRole: cluster-scraper` and a prometheuscr receiver/pipeline in its OTEL config. The per-node agent keeps the default (empty) role.
- Render the TA `ClusterRole`/`ClusterRoleBinding` **once** (all TAs share the SA `target-allocator-service-acct`), so a second TA-bearing agent no longer produces a duplicate-named object.
- Restore the monitor-type disable sentinel selectors; per-agent `allocationStrategy` (targetAgent per-node, clusterScraper consistent-hashing). Extend the CRD (`scraperRole`) and the render-test matrix.

## Note: separate commit
This branch also contains `fix(observability): use non-hyphenated operator feature-gate identifiers` — needed on this base for the operator to start. It is likely a **no-op on current `main`** (main already uses the non-hyphenated IDs); drop that commit if this PR's base already has them.

## Testing
- `helm lint` clean; `tests/prometheus_crds_matrix.sh` 23 pass / 1 pre-existing fail (the CRD-RBAC line belongs to a separate CRD-watch RBAC change not on this base). So it goes green when the base also contains #327's CRD-watch RBAC (i.e. once both the helm stack and #327 land on main). On this isolated stack it can't pass.
- Live-verified: annotated PodMonitor scraped only by the cluster-scraper TA; unannotated only by the per-node TA.

## Dependencies / stacking
Stacks on the helm scraping stack (top = #332, `helm-flag-restructure`). Companion operator PR: aws/amazon-cloudwatch-agent-operator#399.


musa-asad (dismissed):

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -329,6 +346,16 @@ processors:
           - set(attributes["cloudwatch.solution"], "k8s-otel-container-insights")
           - set(attributes["cloudwatch.pipeline"], "efa")
 
+  transform/cw_k8s_ci_v0_set_scope_prometheuscr:
```
musa-asad, line 350: The `prometheuscr` processor is declared in `processors:` but only its pipeline is behind the `prometheusScrape` gate, so with scrape off it renders with zero pipelines and the collector rejects the unused component and won't start. Could we wrap the processor in that same gate?

charts/amazon-cloudwatch-observability/crds/cloudwatch.aws.amazon.com_amazoncloudwatchagents.yaml
```diff
@@ -6576,6 +6577,13 @@ spec:
                           Default: "30s"
                         format: duration
                         type: string
+                      scraperRole:
```
musa-asad, line 6580: `scraperRole` and the new `per-node` `allocationStrategy` only land in the `crds/` copy Helm doesn't update on upgrade, so the API server prunes them from the rendered CR and routing silently fails. Could we move this CRD into `templates/` with `helm.sh/resource-policy: keep`?

charts/amazon-cloudwatch-observability/templates/prometheus-crds/podmonitor-crd.yaml
```diff
@@ -0,0 +1,1429 @@
...
+{{- end }}
+{{- if $render }}
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    helm.sh/resource-policy: keep
+    controller-gen.kubebuilder.io/version: v0.20.1
+    operator.prometheus.io/version: 0.91.0
```
musa-asad: This CRD pins `operator.prometheus.io/version: 0.91.0` but the operator go.mod is on `v0.92.0`, so the API server prunes any monitor field added between the two. Could we vendor both CRDs at `v0.92.0` to match? Same in `servicemonitor-crd.yaml`.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
```diff
@@ -48,6 +48,23 @@ receivers:
                 - {{ include "kube-state-metrics.name" . }}.{{ .Release.Namespace }}.svc:{{ .Values.kubeStateMetrics.service.port }}
 {{- end }}
 
+{{- if dig "prometheusScrape" "enabled" true .Values.otelContainerInsights }}
+  # ServiceMonitor/PodMonitor scraping routed to the cluster-scraper via the
+  # cloudwatch.aws/scraper: cluster-scraper label. This agent's Target Allocator serves the
```
musa-asad: This comment says routing goes by the `cloudwatch.aws/scraper: cluster-scraper` label, but the operator reads it as an annotation, so anyone who follows it sets a label and gets no routing. Could we say annotation here to match the others?

charts/amazon-cloudwatch-observability/tests/prometheus_crds_matrix.sh
```diff
@@ -0,0 +1,109 @@
...
+# always: bundle even with otelCI off.
+expect_count "always + otelCI off => ServiceMonitor CRD bundled" "$SM_CRD" 1 --set otelContainerInsights.enabled=false --set otelContainerInsights.prometheusScrape.crds.install=always
+expect_count "always + otelCI off => PodMonitor CRD bundled"     "$PM_CRD" 1 --set otelContainerInsights.enabled=false --set otelContainerInsights.prometheusScrape.crds.install=always
+
+printf "\n${Y}== resource-policy keep on bundled CRDs ==${N}\n"
+expect_count "bundled CRDs carry resource-policy keep" 'helm.sh/resource-policy: keep' 2 --set otelContainerInsights.enabled=true
+
+printf "\n${Y}== Target Allocator CRD RBAC ==${N}\n"
+# The TA needs customresourcedefinitions get;list;watch on the prometheus-CR path.
+expect_count "CRD RBAC present when otelCI on"       "$CRD_RBAC" 1 --set otelContainerInsights.enabled=true
```
musa-asad, line 97: The CRD RBAC assertions on lines 76 to 78 expect `customresourcedefinitions` RBAC from a separate branch, so this branch renders zero and `set -euo pipefail` fails the suite. Could we move them into that branch or guard them until the rule exists?

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
```diff
@@ -281,6 +322,12 @@ spec:
   {{- if $.Values.otelContainerInsights.enabled }}
```
musa-asad, line 323: `POD_NAME` is injected under the `otelContainerInsights.enabled` guard, so every agent gets it even though only the scraper agents' `prometheusCR` receivers use `${env:POD_NAME}`. It's harmless, but could we scope the guard to `$otelCIScrape` so it lives where it's consumed?

charts/amazon-cloudwatch-observability/templates/target-allocator-clusterrole.yaml
```diff
@@ -1,6 +1,22 @@
...
+*/}}
+{{- $needsTA := false }}
+{{- $needsCR := false }}
 {{- range $i, $customAgent := .Values.agents }}
-{{- if and (and (hasKey ($customAgent.prometheus) "targetAllocator") (hasKey ($customAgent.prometheus.targetAllocator) "enabled")) $customAgent.prometheus.targetAllocator.enabled }}
+{{- $otelCIScrape := eq (include "cloudwatch-agent.otelCIScrapeEnabled" (dict "agentName" $customAgent.name "context" $)) "true" }}
+{{- if or (and (hasKey ($customAgent.prometheus) "targetAllocator") (hasKey ($customAgent.prometheus.targetAllocator) "enabled") $customAgent.prometheus.targetAllocator.enabled) $otelCIScrape }}
+{{- $needsTA = true }}
+{{- end }}
+{{- if or (and (hasKey ($customAgent.prometheus) "targetAllocator") (hasKey ($customAgent.prometheus.targetAllocator) "prometheusCR") $customAgent.prometheus.targetAllocator.prometheusCR.enabled) $otelCIScrape }}
```
musa-asad: `$needsCR` keys on `prometheusCR.enabled` or `$otelCIScrape` but, unlike `$needsTA`, doesn't also require `targetAllocator.enabled`, so a stale `prometheusCR.enabled` on a disabled agent over grants the `monitoring.coreos.com` read rule. Could we mirror `$needsTA` here to keep it least privilege?

charts/amazon-cloudwatch-observability/templates/_helpers.tpl
```diff
@@ -191,6 +210,22 @@ Logic:
...
+applies to the given agent. True when otelContainerInsights is enabled,
+prometheusScrape is enabled, and the agent is either the configured targetAgent
+(per-node scraping of unrouted monitors) or the clusterScraperAgent (central
+scraping of monitors explicitly routed with cloudwatch.aws/scraper: cluster-scraper).
+Accepts a dict with "agentName" (string) and "context" (root context $).
+*/}}
+{{- define "cloudwatch-agent.otelCIScrapeEnabled" -}}
+{{- $ctx := .context -}}
+{{- $agentName := .agentName -}}
+{{- if and $ctx.Values.otelContainerInsights.enabled (dig "prometheusScrape" "enabled" true $ctx.Values.otelContainerInsights) (or (eq $agentName $ctx.Values.otelContainerInsights.targetAgent) (eq $agentName $ctx.Values.otelContainerInsights.clusterScraperAgent)) -}}
```
musa-asad, line 232: `otelCIScrapeEnabled` keys on `prometheusScrape.enabled` alone, so with `serviceMonitor.enabled=false` and `podMonitor.enabled=false` it still renders an idle `target-allocator` that bundles CRDs but discovers nothing. Could we short circuit or reject that both false case?

charts/amazon-cloudwatch-observability/templates/target-allocator-clusterrole.yaml
```diff
@@ -22,12 +38,10 @@ rules:
     verbs: ["get", "list", "watch"]
   - nonResourceURLs: ["/metrics"]
     verbs: ["get"]
-  {{- if and (hasKey ($customAgent.prometheus.targetAllocator) "prometheusCR") $customAgent.prometheus.targetAllocator.prometheusCR.enabled }}
+  {{- if $needsCR }}
   - apiGroups: [ "monitoring.coreos.com"]
```
musa-asad, line 47: This branch's TA `ClusterRole` has the SM/PM rule under `$needsCR` but not the `customresourcedefinitions` rule, which is in #327, so #327 needs to land alongside this for the `target-allocator` to list CRDs. Could we make sure they merge together so discovery can start?

musa-asad (dismissed):

Aakash-Dantre:

## PR Checklist

- [x] Breaking or customer-visible changes are called out in the PR description

> Notes for the description: (1) the bundled `AmazonCloudWatchAgent` CRD now carries
> `enum: [cluster-scraper]` on `scraperRole`, matching the operator-generated schema — previously the
> chart shipped it as a bare string so the constraint was unenforced on-cluster. (2) The commit
> `fix(observability): use non-hyphenated operator feature-gate identifiers` is unrelated to
> annotation routing and would be cleaner as its own PR.

PR #325, opened 2026-07-02
Add Karpenter metrics integration to OTEL Container Insights

vaishnavi-30-beep:
Supersedes #322  (retargeted to release-6.3.0)
This PR adds a Karpenter metrics integration to the OTEL Container Insights pipeline. 
When enabled, the cluster-scraper agent automatically discovers and scrapes Karpenter's 
Prometheus metrics endpoint, enriches them with Kubernetes semantic conventions, and 
exports to CloudWatch via OTLP.

*Description of changes:*
**values.yaml**
- Added `otelContainerInsights.solutions.karpenter` configuration section with 
  `enabled`, `namespace` parameters

**_otel-container-insights-cluster-scraper-config.tpl**
- Added Prometheus receiver using Kubernetes pod service discovery
- Added scope processor for pipeline attribution
- Added groupbyattrs and transform processors for K8s semantic convention enrichment
- Added dedicated resourcedetection instance (cloud-level attributes only, no host/AZ)
- Added pipeline definition wiring receiver through processors to exporter

## Testing

- helm template renders correctly for all flag combinations
- All existing test cases pass (flag_matrix.sh, minikube scenarios)
- Deployed and verified on EKS cluster with Karpenter running
- Metrics confirmed visible in CloudWatch via PromQL
- Integration test PR: https://github.com/aws/amazon-cloudwatch-agent-test/pull/715

CI integration tests passing for Karpenter under eks_daemon:otel_standard_test: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/28984147604/job/86009444350
Screenshot 2026-07-09 at 10 58 37
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #353, opened 2026-08-07
feat(otel-ci): add LLM model-serving telemetry (vLLM, KServe, Knative) as Container Insights solutions

Aakash-Dantre:
# feat(otel-ci): add LLM model-serving telemetry (vLLM, KServe, Knative) as Container Insights solutions

## What

Adds opt-in telemetry for the **five independent sources** a KServe LLM deployment
emits — four Prometheus scrapes and one OTLP trace receiver — wired into the OTEL
Container Insights pipeline as `otelContainerInsights.solutions.*` entries. All
default to **off**.

| Toggle | Source | Collected by | Emits | Destination |
|---|---|---|---|---|
| `solutions.vllm` | vLLM engine, model container | daemonset (per node) | `vllm:*` engine SLIs + `http_*` request counts | CloudWatch metrics |
| `solutions.vllm.traces` | vLLM engine's own OTLP exporter | daemonset (OTLP receiver) | one `llm_request` span per inference, `gen_ai.*` attributes | **CloudWatch OTLP traces endpoint** |
| `solutions.kserve.controlPlane` | `kserve-controller-manager` | cluster-scraper | operator reconcile health | CloudWatch metrics |
| `solutions.knative.dataPlane` | `queue-proxy` sidecar | daemonset (per node) | per-request rate/latency at the proxy | CloudWatch metrics |
| `solutions.knative.controlPlane` | `autoscaler` / `activator` / `controller` / `webhook` | cluster-scraper | scaling decisions, reconcile health | CloudWatch metrics |

The four metric paths follow the established pattern (`ebs-csi-node`, `lis-csi-node`,
Karpenter #325, KEDA): a `prometheus/*` receiver + keep-filter + scope transform + a
dedicated `metrics/*` pipeline exporting to the shared CloudWatch exporter. Traces are
the one exception — they are *pushed* to the agent, not scraped, and leave via an
`otlphttp` exporter to `xray.<region>.amazonaws.com/v1/traces` rather than the metrics
exporter.

> **Metric names depend on the Knative Serving version.** Serving **1.19** moved
> observability to the OpenTelemetry SDK and renamed everything it emits
> (`autoscaler_desired_pods` → `kn_revision_pods_desired`). Both name sets are
> handled; the lists below are labelled by version. See
> [Knative 1.19 and the metric rename](#knative-119-and-the-metric-rename).

## Why four toggles, and why they're named this way

This is the part worth reviewing carefully, because the naming is deliberate and an
earlier revision of this PR got it wrong.

**KServe is a control plane, not a data plane.** It reconciles an `InferenceService` CR
into runtime objects and then steps out of the request path entirely. It never sees an
inference request. So "KServe request telemetry" does not exist — only KServe *operator*
telemetry. Verified against a live controller: its `/metrics` exposes 57 metric families
and **zero** of them are inference metrics. They are all generic `controller_runtime_*`
/ `workqueue_*` / `rest_client_*` / `leader_election_*` families, with KServe specificity
confined to the `controller` label *value* (`inferenceservice`, `inferencegraph`,
`trainedmodel`).

**What KServe reconciles into depends on its deployment mode**, and that determines which
sources exist at all:

```
you apply:   InferenceService  (serving.kserve.io)
                    │
      KServe controller reconciles it, then leaves the request path
                    │
   ┌────────────────┴─────────────────┐
   │ Serverless (KServe's default)    │ RawDeployment
   │  → Knative Service               │  → Deployment + Service + HPA
   │    → Configuration → Revision    │
   │      → Deployment                │     Knative not involved at all
   │        ├─ kserve-container       │     ├─ kserve-container
   │        └─ queue-proxy  ◄─────────┼─────┘  (no sidecar)
   └──────────────────────────────────┘
```

Mode is set cluster-wide via the `inferenceservice-config` ConfigMap
(`{"defaultDeploymentMode": "Serverless"}`) and can be overridden per-InferenceService
with the `serving.kserve.io/deploymentMode` annotation, so both modes can coexist in one
cluster.

Consequently the **request-level metrics come from Knative's queue-proxy, not from
KServe** — hence `knative.dataPlane` rather than `kserve.dataPlane`. Naming that
scrape after KServe would have been actively misleading: it keys on the
`serving.knative.dev/revision` pod label, so it also covers plain Knative Services with
no KServe involvement.

Coverage by mode:

| | Serverless | RawDeployment |
|---|---|---|
| `solutions.vllm` | ✅ | ✅ |
| `solutions.vllm.traces` | ✅ | ✅ |
| `solutions.kserve.controlPlane` | ✅ | ✅ |
| `solutions.knative.dataPlane` | ✅ | ❌ no queue-proxy exists |
| `solutions.knative.controlPlane` | ✅ | ❌ namespace absent |

In RawDeployment the autoscaling view comes from `kube_horizontalpodautoscaler_*` via
kube-state-metrics (already on by chart default), or from the existing `solutions.keda`
toggle — KEDA being the common choice for LLM autoscaling, since CPU/memory is a poor
proxy for GPU inference load. Splitting the toggles this way means a RawDeployment user
enables two (or three, with KEDA) and never renders a Knative scrape that would find no
targets.

Both control-plane scrapes live on the **cluster-scraper**, not the daemonset, because
both components are cluster-wide singletons rather than per-node.

## Metrics emitted

Every metric below was enumerated from live CloudWatch telemetry by scanning every metric
name in the test account and attributing each to its `cloudwatch.pipeline` attribute. Two
clusters were inventoried, one on each side of the Knative rename — 1009 names on the
Knative 1.15 cluster, 943 on the Knative 1.21.1 cluster.

| Toggle | Cluster A — Knative 1.15, KServe 0.14 | Cluster B — Knative 1.21.1, KServe 0.15 |
|---|---|---|
| `solutions.vllm` | 39 | 39 — identical set |
| `solutions.kserve.controlPlane` | 17 | 21 — 4 extra `controller_runtime_*` families from a newer controller-runtime |
| `solutions.knative.dataPlane` | 4 | 2 — renamed |
| `solutions.knative.controlPlane` | 52 | 29 — renamed |

The two Knative versions share **no** metric names at all. The KServe delta is a
controller-runtime version difference, not a rename.

<details>
<summary><b><code>solutions.vllm</code> → 39 names</b></summary>

Engine (`vllm:*`, 36) — latency and throughput SLIs, cache behaviour, scheduler state:

```
vllm:time_to_first_token_seconds          vllm:inter_token_latency_seconds
vllm:e2e_request_latency_seconds          vllm:request_time_per_output_token_seconds
vllm:request_queue_time_seconds           vllm:request_prefill_time_seconds
vllm:request_decode_time_seconds          vllm:request_inference_time_seconds
vllm:num_requests_running                 vllm:num_requests_waiting
vllm:num_requests_waiting_by_reason       vllm:num_preemptions_total
vllm:kv_cache_usage_perc                  vllm:prefix_cache_hits_total
vllm:prefix_cache_queries_total           vllm:external_prefix_cache_hits_total
vllm:external_prefix_cache_queries_total  vllm:mm_cache_hits_total
vllm:mm_cache_queries_total               vllm:prompt_tokens_total
vllm:generation_tokens_total              vllm:iteration_tokens_total
vllm:prompt_tokens_cached_total           vllm:prompt_tokens_by_source_total
vllm:request_prompt_tokens                vllm:request_generation_tokens
vllm:request_max_num_generation_tokens    vllm:request_prefill_kv_computed_tokens
vllm:request_params_max_tokens            vllm:request_params_n
vllm:request_success_total                vllm:cache_config_info
vllm:engine_sleep_state                   vllm:estimated_flops_per_gpu_total
vllm:estimated_read_bytes_per_gpu_total   vllm:estimated_write_bytes_per_gpu_total
```

HTTP (`http_*`, 3) — from the FastAPI instrumentator, labelled `handler` / `method` /
`status`:

```
http_requests_total   http_request_duration_seconds   http_request_duration_highr_seconds
```

These are kept deliberately: they are the only HTTP-status telemetry vLLM emits, which
makes them the stand-in for `revision_*` in RawDeployment mode where there is no
queue-proxy.

Every series above also carries an injected `inferenceservice` label, the KServe CR name.

</details>

<details>
<summary><b><code>solutions.kserve.controlPlane</code> → 17–21 names</b></summary>

Operator health only — whether `InferenceService` reconciliation is succeeding, not how
models are serving. 17 of the controller's 57 families survive the keep-filter on KServe
0.14:

```
controller_runtime_reconcile_total              controller_runtime_reconcile_errors_total
controller_runtime_reconcile_time_seconds       controller_runtime_active_workers
controller_runtime_max_concurrent_reconciles    controller_runtime_webhook_requests_total
controller_runtime_webhook_latency_seconds      controller_runtime_webhook_requests_in_flight
leader_election_master_status                   rest_client_requests_total
workqueue_depth                                 workqueue_adds_total
workqueue_queue_duration_seconds                workqueue_work_duration_seconds
workqueue_retries_total                         workqueue_unfinished_work_seconds
workqueue_longest_running_processor_seconds
```

KServe 0.15 ships a newer controller-runtime, which adds 4 more — all already matched by
the existing `^controller_runtime_` keep-pattern, so no filter change is needed:

```
controller_runtime_reconcile_panics_total       controller_runtime_webhook_panics_total
controller_runtime_conversion_webhook_panics_total
controller_runtime_terminal_reconcile_errors_total
```

</details>

<details>
<summary><b><code>solutions.knative.dataPlane</code> → 4 names (≤ 1.18) / 2 names (≥ 1.19)</b></summary>

**Knative Serving ≤ 1.18**

```
revision_request_count          revision_request_latencies
revision_app_request_count      revision_app_request_latencies
```

`revision_*` is measured at the queue-proxy and therefore includes queue wait time;
`revision_app_*` is measured at the user container. Both are labelled `response_code`,
`configuration_name`, and (injected by this PR) `inferenceservice`. The difference
between the two is the queueing overhead Knative adds.

**Knative Serving ≥ 1.19** — `pkg/queue/request_metric.go` now declares two OTel
instruments on scope `knative.dev/serving/pkg/queue`, which the SDK's Prometheus
exporter renders as:

```
kn_serving_invocation_duration_seconds   (histogram, from kn.serving.invocation.duration)
kn_serving_queue_depth                   (gauge,     from kn.serving.queue.depth)
```

The proxy-vs-app split is gone; the histogram carries
`http.response.status_code` (semconv) in place of `response_code`.

**This is the one part of the inventory not confirmed against live CloudWatch data**, for
the reason in the next section: on the 1.21.1 cluster the queue-proxy was not exporting
request metrics at all, so the names above are read from the 1.21 source rather than
observed. The keep-filter is therefore written as a prefix (`^kn_serving_.*`) rather than
an exact list, so it holds regardless of how the exporter suffixes the unit.

</details>

<details>
<summary><b><code>solutions.knative.controlPlane</code> → 52 names (≤ 1.18) / 29 names (≥ 1.19)</b></summary>

### Knative Serving ≥ 1.19

Observed on the 1.21.1 cluster: the scrape yields 66 names, of which 29 survive the
keep-filter. **The scaling decision (24, `kn_*`)** — same information as the old
`autoscaler_*` group, renamed and re-grouped around the revision:

```
kn_revision_pods_desired              kn_revision_pods_count
kn_revision_pods_requested            kn_revision_pods_pending_count
kn_revision_pods_not_ready_count      kn_revision_pods_terminating_count
kn_revision_concurrency_stable        kn_revision_concurrency_panic
kn_revision_concurrency_target        kn_revision_request_concurrency
kn_revision_panic_mode                kn_revision_capacity_excess
kn_activator_stats_conn_errors_total  kn_activator_stats_conn_reachable
kn_autoscaler_scrape_duration_seconds kn_webhook_handler_duration_seconds
kn_k8s_client_http_response_status_code_total
kn_workqueue_depth                    kn_workqueue_adds_total
kn_workqueue_retries_total            kn_workqueue_queue_duration_seconds
kn_workqueue_process_duration_seconds kn_workqueue_unfinished_work_seconds
kn_workqueue_longest_running_processor_seconds
```

**HTTP (5)** — the OTel SDK replaces the old bespoke `webhook_request_count` /
`activator_request_count` families with OTel HTTP semantic conventions:

```
http_server_request_duration_seconds   http_server_request_body_size_bytes
http_server_response_body_size_bytes   http_client_request_duration_seconds
http_client_request_body_size_bytes
```

Mapping for anyone porting a dashboard:

| ≤ 1.18 | ≥ 1.19 |
|---|---|
| `autoscaler_desired_pods` | `kn_revision_pods_desired` |
| `autoscaler_actual_pods` | `kn_revision_pods_count` |
| `autoscaler_requested_pods` | `kn_revision_pods_requested` |
| `autoscaler_not_ready_pods` / `_pending_pods` / `_terminating_pods` | `kn_revision_pods_not_ready_count` / `_pending_count` / `_terminating_count` |
| `autoscaler_stable_request_concurrency` | `kn_revision_concurrency_stable` |
| `autoscaler_panic_request_concurrency` | `kn_revision_concurrency_panic` |
| `autoscaler_target_concurrency_per_pod` | `kn_revision_concurrency_target` |
| `autoscaler_panic_mode` | `kn_revision_panic_mode` |
| `autoscaler_excess_burst_capacity` | `kn_revision_capacity_excess` |
| `autoscaler_scrape_time` | `kn_autoscaler_scrape_duration_seconds` |
| `webhook_request_count` / `_latencies` | `http_server_request_duration_seconds`, `kn_webhook_handler_duration_seconds` |
| `*_client_latency` / `_client_results` | `http_client_request_duration_seconds`, `kn_k8s_client_http_response_status_code_total` |
| `*_workqueue_*` / `*_work_queue_depth` | `kn_workqueue_*` (single set — the legacy/current duplication is gone) |

Series identity moved too: `revision_name` / `configuration_name` / `service_name` became
`kn_revision_name` / `kn_configuration_name` / `kn_service_name`, and the scope is now
reported as `otel_scope_name="knative.dev/serving/pkg/autoscaler"`.

### Knative Serving ≤ 1.18

**autoscaler (24)** — the scaling decision, the most operationally useful group here:

```
autoscaler_desired_pods              autoscaler_actual_pods
autoscaler_requested_pods            autoscaler_pending_pods
autoscaler_not_ready_pods            autoscaler_terminating_pods
autoscaler_stable_request_concurrency autoscaler_panic_request_concurrency
autoscaler_panic_mode                autoscaler_excess_burst_capacity
autoscaler_target_concurrency_per_pod autoscaler_scrape_time
autoscaler_reconcile_count           autoscaler_reconcile_latency
autoscaler_client_latency            autoscaler_client_results
+ 8 autoscaler_workqueue_* / work_queue_depth
```

**webhook (14)** — `webhook_request_count`, `webhook_request_latencies`, plus
reconcile/workqueue families.

**controller (12)** — `controller_reconcile_count`, `controller_reconcile_latency`,
`controller_client_latency`, `controller_client_results`, plus workqueue families.

**activator (2)** — `activator_client_latency`, `activator_client_results`. Note that
`activator_request_count` / `_latencies` are absent unless a scale-from-zero cold start
occurs, since the activator is only in the data path while a revision is scaled to zero.

Knative emits both legacy and current workqueue names (`autoscaler_work_queue_depth`
*and* `autoscaler_workqueue_depth`) — duplicate data from upstream, not introduced here.

</details>

## Keep-filters

Every pipeline filters at the collector so that only relevant families reach CloudWatch,
rather than everything the endpoint happens to expose:

| Pipeline | Filter |
|---|---|
| `vllm` | keep `^vllm:` and `^http_`; drop `*_created` |
| `kserve-controlplane` | keep `controller_runtime_` / `workqueue_` / `rest_client_` / `leader_election_` |
| `knative-dataplane` | keep `^revision_.*request.*` (≤ 1.18) and `^kn_serving_.*` (≥ 1.19) |
| `knative-controlplane` | drop `*_go_*`, `^go_*`, `^process_*`, `^promhttp_*` |

The vLLM filter is worth calling out because the endpoint exposes considerably more than
the engine's own series. Measured on the test deployment, it went from **84 metric names
to 39**. Dropped:

- **`*_created` (35 names)** — `client_python` emits a creation-timestamp companion gauge
  for every counter and histogram. The value is a constant unix timestamp, identical on
  every scrape for the life of the process. Their one legitimate use is counter-reset
  detection, which the pipeline already handles via `metricstarttime`; dropping them in a
  processor is safe because the Prometheus receiver has already consumed them to set start
  timestamps before processors run.
- **`process_*` (6)** — API-server process CPU/memory/fd stats. Already covered at the
  container level by Container Insights core, and colliding by metric name with the
  kube-apiserver's own `process_*` series, which makes the metric ambiguous on a dashboard.
- **`python_*` (4)** — interpreter GC counters plus a constant-`1` `python_info` gauge.

## Knative 1.19 and the metric rename

Knative Serving **1.19** replaced its OpenCensus metrics with the OpenTelemetry SDK. Every
metric it emits was renamed, all of them gaining a `kn` prefix. Re-testing this feature on
a **1.21.1** cluster surfaced three consequences, all fixed in this PR:

1. **`knative.dataPlane` delivered nothing.** The keep-filter was `^revision_.*request.*`,
   and no `revision_*` series exist on ≥ 1.19 — the whole toggle was a no-op, with a valid
   config and no error anywhere. Now keeps `^kn_serving_.*` as well.
2. **The control-plane runtime filter stopped matching.** It drops `.*_go_.*`, which fits
   the pre-1.19 per-component names (`autoscaler_go_gc_duration_seconds`). On ≥ 1.19 the
   Go runtime metrics are bare `go_*`, and `.*_go_.*` does not match `go_*` — it requires a
   leading underscore. Result: 35 `go_*` families plus 2 `promhttp_*` were billed into
   CloudWatch, more than doubling this pipeline's volume. Both patterns are now listed.
   This also corrects a claim in the earlier revision of this description: the check that
   "`revision_go_*` yields 0 series" was true but not sufficient, because it only proved
   the pre-1.19 name was absent.
3. **The data plane needs a second opt-in inside Knative.** On ≥ 1.19 the queue-proxy's
   request metrics are gated by their own key, `request-metrics-protocol` in the
   `config-observability` ConfigMap, which defaults to **none**. `metrics-protocol`
   governs the control plane only. While it is none, port 9091 (`http-usermetric`) refuses
   connections outright, so the scrape yields nothing no matter how the chart is
   configured:

   ```bash
   kubectl -n knative-serving patch cm config-observability --type merge \
     -p '{"data":{"request-metrics-protocol":"prometheus"}}'
   # queue-proxy reads this from injected env, so existing pods must be recycled
   kubectl -n <ns> rollout restart deploy <revision>-deployment
   ```

   This is now documented on the `knative.dataPlane` values block, since without it the
   toggle looks broken.

Both name sets are kept rather than switching on a version, so one chart covers clusters
either side of 1.19 and there is no version to detect at render time.

## vLLM request traces (`solutions.vllm.traces`)

vLLM's engine emits **one `llm_request` span per inference** over OTLP — its own spans, not
synthesized here — carrying the `gen_ai.*` latency and token-count attributes. Enabling
this opens an OTLP receiver on the agent and forwards those spans to the **CloudWatch OTLP
traces endpoint**, giving per-request drill-down that metrics cannot: the latency breakdown
of a *single* slow request, rather than a percentile over all of them.

```
vLLM engine  ──OTLP──▶  cloudwatch-agent (daemonset)  ──OTLP/HTTP──▶  xray.<region>.amazonaws.com/v1/traces
  --otlp-traces-endpoint      otlp receiver :4319 (gRPC) / :4320 (HTTP)
                              k8sattributes → resource transform
                              → resourcedetection → batch
                              → otlphttp exporter, SigV4 (service: xray)
```

Push, not scrape, so unlike the four metric toggles enabling it does nothing on its own —
the engine sends nothing until it is started with an endpoint:

```yaml
spec:
  predictor:
    containers:
      - name: kserve-container
        args:
          - --otlp-traces-endpoint=grpc://cloudwatch-agent.amazon-cloudwatch:4319
```

`grpc://` or `http://`, never `https://` — the scheme selects TLS, not the protocol, and
the receiver serves plaintext. (vLLM's exporter is OTLP/gRPC unless
`OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` says otherwise.) On vLLM 0.11 also set
`OTEL_EXPORTER_OTLP_TRACES_INSECURE=true`; 0.26 needs nothing.

**Ports.** 4319/4320 rather than the conventional 4317/4318, because the agent runs with
`hostNetwork: true` and 4317/4318 are what a customer's own collector is most likely to
hold — Application Signals avoids them for the same reason, using 4315/4316. Both are
configurable.

**Why the OTLP endpoint and not the `awsxray` exporter.** An earlier revision of this PR
used `awsxray`. That exporter converts spans into **X-Ray segments**, which is a lossy
projection in three ways: only attributes named in an explicit `indexed_attributes` list
become filterable (X-Ray caps annotations at 50 per segment, so `index_all_attributes` is
not a safe default), dotted keys are subject to the `exporter.xray.allowDot` feature gate
so the searchable name may or may not match what vLLM emitted, and trace IDs are
X-Ray-format rather than W3C.

The OTLP endpoint stores spans in the **OpenTelemetry semantic-convention format** with W3C
trace IDs, so every attribute vLLM sets — the whole `gen_ai.*` set plus the resource
attributes this pipeline adds — stays queryable in Transaction Search. There is no indexed
subset to declare and the 16-entry `indexed_attributes` list is gone.

To be precise about one thing that is easy to get wrong: this is **not** about reachability.
X-Ray segments also reach Transaction Search — they are automatically converted to the
semantic-convention format before being stored in the `aws/spans` log group. The gain here
is fidelity and one less config surface, not access to a destination that was otherwise
closed.

Endpoint details worth knowing:

- It is **HTTP only** — it does not accept gRPC — hence `otlphttp`, not `otlp`. This is
  independent of the vLLM→agent hop, which stays gRPC because that is vLLM's default
  (`vllm/tracing.py` reads `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` defaulting to `"grpc"`).
  The receiver opens both ports regardless.
- Auth is SigV4 with signing name `xray`, via the `sigv4auth` extension. Bearer tokens are
  not supported for traces.
- Batches are 50 spans, well inside the endpoint's 10,000-span / 5 MB-uncompressed
  per-request limits. Compression is `gzip`.
- **Transaction Search must be enabled on the account.** Until it is, the endpoint is not
  usable and spans do not appear. This is a new prerequisite relative to `awsxray` and is
  documented on the values flag.

The pipeline also fills in identity the engine does not send: `k8sattributes` (node-scoped,
associated `from: connection`) adds pod, namespace and workload; a transform sets
`k8s.cluster.name` and falls back through inferenceservice → deployment → pod for
`service.name`, so spans are never filed under `unknown_service`.

**Caveats worth reading before enabling this:**

- **Transaction Search must be enabled on the account**, or the endpoint accepts nothing.
  This is an account-wide, per-Region change — see the chart README.
- **Billed per span, and vLLM does no sampling of its own** — one span per inference. On a
  high-QPS endpoint set `OTEL_TRACES_SAMPLER` on the engine first.
- **The endpoint enrols these spans in Application Signals.** It injects `aws.local.*` and
  `aws.span.kind` server-side, which creates a service entity and emits billed
  `ApplicationSignals` metrics (`Latency`, `Error`, `Fault`, `Throttle`, `InputTokens`,
  `OutputTokens`). This is not configured by the chart and cannot be disabled from it. The
  transform sets `deployment.environment` so the entity derives as
  `eks:<cluster>/<namespace>` rather than the `eks:default` it otherwise falls back to.
- **Errors are not observable, and the AppSignals view will look healthy regardless.** vLLM
  never sets span status, and only successfully finished requests are traced at all — aborts,
  timeouts and errors emit no span — so `Error` and `Fault` stay at zero by construction.
  Upstream fix: [vllm#32162](https://github.com/vllm-project/vllm/pull/32162), open since
  January.
- **No model name on the spans.** The V1 engine never sets `gen_ai.response.model` despite
  defining the constant, so spans cannot be joined to the `vllm:*` metrics by model —
  those do carry a `model_name` label. `service.name` carries `--served-model-name`.
- **Traces are single-span.** vLLM honours an inbound W3C `traceparent`, but nothing in a
  default Istio/Knative path propagates one, so there is no end-to-end trace and gateway
  queueing time is invisible.

## Notable fix included

The vLLM scrape originally kept targets on the container **port name** `user-port`. That
name is injected by **Knative**, not KServe — neither the `ServingRuntime` nor the Knative
`Revision` declares a name for the model container's port:

```
ServingRuntime  ports=[{"containerPort":8080,"protocol":"TCP"}]   ← no name
Revision        ports=[{"containerPort":8080,"protocol":"TCP"}]   ← no name
running pod     kserve-container -> user-port:8080                ← name appears
```

Knative's revision reconciler renames the user container's port when it injects the
queue-proxy. In RawDeployment mode there is no queue-proxy and no rename, so the port
stays unnamed, the keep action matches nothing, and the pipeline silently produces no
metrics — valid config, no error, no data.

Now keyed on the container **name** (`kserve-container`), which KServe sets identically in
both modes. Still selects one target per pod and still excludes the `queue-proxy` and
`agent` sidecars.

## Backwards compatibility

Fully backwards-compatible. Every toggle defaults to `false`, and every block is gated on
both `solutions.enabled` and its own flag, so `solutions.enabled: false` acts as a single
kill switch. Traces are gated on `solutions.enabled` ∧ `vllm.enabled` ∧
`vllm.traces.enabled`, so they cannot be enabled without the metrics scrape they belong to.
Verified by `helm template`:

- chart defaults → renders no vLLM/KServe/Knative receivers, processors, or pipelines
- `solutions.enabled: false` with all individual toggles `true` → still renders nothing
- all enabled → four `prometheus/*` receivers + one `otlp/*` receiver, four keep-filters,
  five scope/resource transforms, five pipelines; the traces one exports to
  `otlphttp/…_traces_dest`, the rest to the shared metrics exporter
- no port set (`--reuse-values` drops the defaults) → the chart `fail`s at render with a
  message naming the flag, rather than emitting a config the operator rejects

`helm lint --set region=<region>` clean. (Bare `helm lint` fails on the pre-existing
Windows fluent-bit `repositoryDomainMap` lookup, unrelated to this change — same on the
base commit.)

## Testing

Deployed end-to-end on two live EKS clusters, one either side of the Knative rename, and
every metric above confirmed queryable in CloudWatch via PromQL.

| | Cluster A | Cluster B |
|---|---|---|
| EKS | v1.31 (`v1.31.14-eks-bca9cf6`), eu-west-1 | v1.33, us-west-2 |
| Nodes | 1 × `g5.2xlarge` (A10G GPU) + 2 × `m6i.large` | 1 × `g6e.2xlarge` (L40S GPU) + 2 × `m6i.large` |
| KServe | v0.14.0, Serverless mode | v0.15, Serverless mode |
| Knative Serving | v1.15.0, `net-istio` | **v1.21.1**, `net-istio` |
| vLLM | `vllm/vllm-openai:v0.26.0` (V1 engine), Qwen2.5-7B-Instruct | same |
| Chart | `amazon-cloudwatch-observability-6.5.1` | 6.6.0 pre-release build |
| Covers | the four metric toggles | the rename fixes + traces |

Verification performed on cluster A:

- **Inventory** — scanned all 1009 metric names present in the account and attributed each
  to its `cloudwatch.pipeline` attribute, confirming the four pipelines emit exactly the
  112 names listed above and nothing else. Dotted OTel-native names were re-probed with
  quoted PromQL-3 selector syntax to confirm none belonged to these pipelines.
- **Filter integrity** — confirmed `revision_go_*` yields 0 series, and that no `go_*` /
  `process_*` / `workqueue_depth` series carry the `knative_component` or `controller`
  labels, i.e. the control-plane filters are not leaking runtime noise. *This check was
  necessary but not sufficient — see the 1.19 section; on ≥ 1.19 the runtime metrics arrive
  under a name it does not cover.*
- **Request path** — drove live inference (`POST /v1/chat/completions`) through the
  queue-proxy and confirmed `revision_request_count{response_code="200"}`,
  `revision_app_request_count`, and `histogram_count(revision_request_latencies)` all
  landed with matching values. Note that with zero traffic the queue-proxy exposes only
  `revision_go_*` — the request families do not exist until a request has flowed, so 0
  series while idle is expected rather than a broken scrape.
- **vLLM filter** — verified by range query across a collector restart that the dropped
  families stop producing datapoints while `vllm:*` and `http_*` continue uninterrupted,
  and that the real counters (`vllm:generation_tokens_total`, `vllm:request_success_total`,
  `vllm:prompt_tokens_total`) are unaffected.

On cluster B (Knative 1.21.1):

- **Re-inventory** — enumerated all 943 metric names in the account and attributed each to
  its `cloudwatch.pipeline`, which is how the rename was found: 0 `revision_*`, 0
  `autoscaler_*`, 0 `activator_*`, 0 `webhook_*`, and 24 `kn_*` in their place. The `vllm`
  pipeline's 39 names came back byte-identical to cluster A, confirming the vLLM inventory
  is not version- or GPU-specific.
- **The `go_*` leak** — the 35 `go_*` and 2 `promhttp_*` families were each confirmed to
  carry `cloudwatch.pipeline="knative-controlplane"`, i.e. they were arriving through this
  feature's pipeline and not from some other scrape.
- **Traces, end to end through the OTLP endpoint** — enabled Transaction Search on the test
  account, drove live inference and confirmed the spans land in the `aws/spans` log group in
  semantic-convention format with W3C trace IDs. All 13 `gen_ai.*` attributes the V1 engine
  emits are present and queryable with no indexed subset, `service.name` derives from the
  InferenceService rather than `unknown_service`, and the full pod/node/instance resource
  set is attached. Zero export errors in the agent log; the traces are also retrievable via
  the X-Ray API and resolve in the console once the W3C ID is rewritten to X-Ray form.
- **The Transaction Search prerequisite** — verified by observing the failure first. With the
  account on `Destination: XRay`, every batch was rejected with `HTTP 400 ... The OTLP API is
  supported with CloudWatch Logs as a Trace Segment Destination`, which the collector treats
  as permanent and drops. A semantic 400 rather than a 403 also confirms the SigV4 signing is
  correct. After switching the destination the same log showed zero such errors.
- **Entity derivation** — confirmed the Application Signals entity re-derives from
  `eks:default` to `eks:<cluster>/<namespace>` once `deployment.environment` is set, and that
  `gen_ai.system` and the current-semconv token attribute names arrive on the spans.
- **Not covered:** the ≥ 1.19 data-plane metric names, because
  `request-metrics-protocol` was `none` on this cluster and enabling it needs a revision
  recycle. The filter is a prefix match for that reason.

## Usage

Serverless mode (KServe's default) — all four sources present:

```yaml
otelContainerInsights:
  enabled: true
  solutions:
    enabled: true
    vllm:
      enabled: true
      namespace: llm-serving      # optional; empty = all namespaces
      traces:
        enabled: true
        grpcPort: 4319
        httpPort: 4320
    kserve:
      controlPlane:
        enabled: true
        namespace: kserve
    knative:
      dataPlane:
        enabled: true
        namespace: llm-serving    # optional; empty = all namespaces
      controlPlane:
        enabled: true
        namespace: knative-serving
```

Use `-f values.yaml` rather than `--reuse-values`, which drops the port defaults.

`vllm.traces` additionally requires Transaction Search enabled on the account, and the
engine started with `--otlp-traces-endpoint` as shown above — the chart only opens the
receiver.

On Knative ≥ 1.19, `dataPlane` additionally needs Knative itself to export the metrics:

```bash
kubectl -n knative-serving patch cm config-observability --type merge \
  -p '{"data":{"request-metrics-protocol":"prometheus"}}'
```

RawDeployment mode — Knative is not installed, so enable only the mode-independent
sources:

```yaml
otelContainerInsights:
  enabled: true
  solutions:
    enabled: true
    vllm:
      enabled: true
      namespace: llm-serving
    kserve:
      controlPlane:
        enabled: true
        namespace: kserve
    keda:
      enabled: true               # autoscaling view, replaces autoscaler_*
```

## RBAC

The KServe control-plane scrape goes through the controller's `kube-rbac-proxy` on `:8443`
(HTTPS + the agent ServiceAccount bearer token), which requires `get` on the `/metrics`
non-resource URL. The chart's existing `cloudwatch-agent-role` already grants this, so no
RBAC changes are needed.

vLLM traces need no additional IAM either: `CloudWatchAgentServerPolicy` already grants the
CloudWatch OTLP traces endpoint. Transaction Search does need to be enabled on the account —
that is an account setting, not a permission.

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
+++ b/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
@@ -120,6 +120,40 @@ receivers:
               regex: metrics
               action: keep
 
+  {{- if .Values.otelContainerInsights.vllm.enabled }}
+  # vLLM model-server metrics (vllm:* Prometheus series). Pod-discovery scrape
+  # on the local node, keyed on the KServe InferenceService label + container
+  # port 8080. Mirrors the ebs-csi-node pattern.
+  prometheus/cw_k8s_ci_v0_vllm:
+    config:
+      scrape_configs:
+        - job_name: vllm
+          scrape_interval: {{ .Values.otelContainerInsights.metricResolution }}
+          scrape_timeout: {{ include "otel-container-insights.scrapeTimeout" . }}
+          metrics_path: /metrics
+          kubernetes_sd_configs:
+            - role: pod
+              {{- if .Values.otelContainerInsights.vllm.namespace }}
+              namespaces:
+                names:
+                  - {{ .Values.otelContainerInsights.vllm.namespace }}
+              {{- end }}
+          relabel_configs:
+            # keep only KServe InferenceService pods (match the label VALUE, like
+            # the ebs-csi job keys on pod_label_app)
+            - source_labels: [__meta_kubernetes_pod_label_serving_kserve_io_inferenceservice]
+              regex: .+
+              action: keep
+            # scrape the vLLM container's metrics port (8080, named "user-port")
+            - source_labels: [__meta_kubernetes_pod_container_port_name]
+              regex: user-port
+              action: keep
+            # expose the InferenceService name as a label for CloudWatch dimensioning
+            - source_labels: [__meta_kubernetes_pod_label_serving_kserve_io_inferenceservice]
+              target_label: inferenceservice
+              action: replace
+  {{- end }}
+
   kubeletstats/cw_k8s_ci_v0:
     auth_type: serviceAccount
     collection_interval: {{ .Values.otelContainerInsights.metricResolution }}
@@ -338,6 +372,16 @@ processors:
           - set(attributes["cloudwatch.source"], "cloudwatch-agent")
           - set(attributes["cloudwatch.solution"], "k8s-otel-container-insights")
           - set(attributes["cloudwatch.pipeline"], "ebs-csi")
+
+  transform/cw_k8s_ci_v0_set_scope_vllm:
+    error_mode: ignore
+    metric_statements:
+      - context: scope
+        statements:
+          - set(scope.schema_url, "")
+          - set(attributes["cloudwatch.source"], "cloudwatch-agent")
+          - set(attributes["cloudwatch.solution"], "k8s-otel-container-insights")
+          - set(attributes["cloudwatch.pipeline"], "vllm")
 
   transform/cw_k8s_ci_v0_set_scope_lis_csi:
     error_mode: ignore
@@ -966,6 +1010,28 @@ service:
         - batch/cw_k8s_ci_v0_metrics_dest
       exporters:
         - otlphttp/cw_k8s_ci_v0_metrics_dest
+
+{{- if .Values.otelContainerInsights.vllm.enabled }}
+    metrics/cw_k8s_ci_v0_vllm:
+      receivers: [prometheus/cw_k8s_ci_v0_vllm]
+      processors:
+        - filter/cw_k8s_ci_v0_scrape_metadata
+        - transform/cw_k8s_ci_v0_set_unit
+        - metricstarttime/cw_k8s_ci_v0
+        - transform/cw_k8s_ci_v0_set_cluster_name
+        - transform/cw_k8s_ci_v0_set_node_name
+        - transform/cw_k8s_ci_v0_promote_node_name
+        - resourcedetection/cw_k8s_ci_v0
+        - transform/cw_k8s_ci_v0_set_cloud_resource_id
+        - k8sattributes/cw_k8s_ci_v0_node
+        - transform/cw_k8s_ci_v0_set_scope_vllm
+        - transform/cw_k8s_ci_v0_clear_schema_url
+        - transform/cw_k8s_ci_v0_set_workload
+        - awsattributelimit/cw_k8s_ci_v0
+        - batch/cw_k8s_ci_v0_metrics_dest
+      exporters:
+        - otlphttp/cw_k8s_ci_v0_metrics_dest
+{{- end }}
 
     metrics/cw_k8s_ci_v0_lis_csi_node:
       receivers: [prometheus/cw_k8s_ci_v0_lis_csi_node]
--- a/charts/amazon-cloudwatch-observability/values.yaml
+++ b/charts/amazon-cloudwatch-observability/values.yaml
@@ -1195,3 +1195,10 @@ otelContainerInsights:
       enabled: true
       ## Namespace where KEDA is deployed.
       namespace: "keda"
+  ## Optional Prometheus scraping of vLLM model-server metrics (vllm:* series)
+  ## from KServe InferenceService pods, published through the OTEL CI pipeline.
+  ## Only takes effect when otelContainerInsights.enabled is true.
+  vllm:
+    enabled: false
+    ## Namespace where the vLLM/KServe InferenceService pods run. Empty = all namespaces.
+    namespace: ""
--- a/charts/amazon-cloudwatch-observability/values.yaml
+++ b/charts/amazon-cloudwatch-observability/values.yaml
@@ -1173,3 +1173,10 @@ otelContainerInsights:
   cloudwatchMetricsEndpoint: ""
   # Override the CloudWatch Logs OTLP endpoint. If empty, defaults to logs.<region>.amazonaws.com.
   cloudwatchLogsEndpoint: ""
+  ## Optional Prometheus scraping of vLLM model-server metrics (vllm:* series)
+  ## from KServe InferenceService pods, published through the OTEL CI pipeline.
+  ## Only takes effect when otelContainerInsights.enabled is true.
+  vllm:
+    enabled: false
+    ## Namespace where the vLLM/KServe InferenceService pods run. Empty = all namespaces.
+    namespace: ""


charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1173,18 +1173,10 @@ otelContainerInsights:
...
-    ## Individual solutions won't activate even if their own enabled flag is true.
-    enabled: true
-    ## Karpenter — cluster autoscaler for Kubernetes.
-    ## Scrapes the Prometheus metrics endpoint exposed by the Karpenter controller.
-    karpenter:
-      enabled: true
-      ## Namespace where Karpenter is deployed.
-      namespace: "kube-system"
+  ## Optional Prometheus scraping of vLLM model-server metrics (vllm:* series)
+  ## from KServe InferenceService pods, published through the OTEL CI pipeline.
```
oberkem, line 1199: why only kserve?

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -120,6 +125,102 @@ receivers:
...
+          scrape_timeout: {{ include "otel-container-insights.scrapeTimeout" . }}
+          metrics_path: /metrics
+          kubernetes_sd_configs:
+            - role: pod
+              {{- if .Values.otelContainerInsights.solutions.vllm.namespace }}
+              namespaces:
+                names:
+                  - {{ .Values.otelContainerInsights.solutions.vllm.namespace }}
+              {{- end }}
+          relabel_configs:
```
oberkem, line 146: don't we need 
```            
- source_labels: [__meta_kubernetes_pod_node_name]
              regex: ${env:K8S_NODE_NAME}
              action: keep
```
like csi-node job?

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -120,6 +125,102 @@ receivers:
...
+  # downstream (see the *_keep filter).
+  #
+  # The port only serves once Knative is told to export request metrics. On
+  # Serving >= 1.19 that is `request-metrics-protocol: prometheus` in the
+  # config-observability ConfigMap; it defaults to none, and while it is none
+  # the port refuses connections and this scrape finds nothing.
+  prometheus/cw_k8s_ci_v0_knative_dataplane:
+    config:
+      scrape_configs:
+        - job_name: knative-dataplane
```
oberkem, line 198: same here https://github.com/aws-observability/helm-charts/pull/353/changes#r4075661295

charts/amazon-cloudwatch-observability/README.md
```diff
@@ -19,6 +19,129 @@ helm install --wait --create-namespace --namespace amazon-cloudwatch amazon-clou
...
+`aws.span.kind` server-side, which creates a service entity and emits billed
+`ApplicationSignals` metrics (`Latency`, `Error`, `Fault`, `Throttle`, `InputTokens`,
+`OutputTokens`). That is not configured by this chart and cannot be turned off from here.
+
+Three limitations come from vLLM itself and are worth knowing before building on this:
+
+- **Errors are not observable.** vLLM never sets span status, and only successfully
+  finished requests are traced at all — aborts, timeouts and errors emit no span. So
+  Application Signals `Error` and `Fault` stay at zero regardless of what the engine is
+  doing. Upstream fix: [vllm#32162](https://github.com/vllm-project/vllm/pull/32162).
```
oberkem, line 100: this PR has been open for 8 months. I asked the author about the status. Can you get that CR to the finish line?

charts/amazon-cloudwatch-observability/README.md
```diff
@@ -19,6 +19,129 @@ helm install --wait --create-namespace --namespace amazon-cloudwatch amazon-clou
...
+          - --otlp-traces-endpoint=grpc://cloudwatch-agent.amazon-cloudwatch:4319
+```
+
+Use `grpc://` or `http://`, never `https://` — the scheme selects TLS, not the protocol, and
+the receiver serves plaintext. vLLM's exporter is OTLP/gRPC unless
+`OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=http/protobuf` is set, in which case use the HTTP port.
+On vLLM 0.11 also set `OTEL_EXPORTER_OTLP_TRACES_INSECURE=true`; 0.26 needs nothing.
+`--collect-detailed-traces` is not needed — it does nothing for traces on the V1 engine.
+
+Spans are billed per trace and vLLM emits one per request with no sampling of its own, so on
```
oberkem, line 127: instead of disabling the whole thing, should we enable it with 10% sampling rate

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
```diff
@@ -375,6 +403,58 @@ processors:
     {{- end }}
 {{- end }}
 
+{{- if and .Values.otelContainerInsights.solutions.enabled .Values.otelContainerInsights.solutions.kserve.controlPlane.enabled }}
+  # Keep only the KServe operator health families; drop go_*/process_* runtime noise.
```
oberkem, line 407: you can do the same to vllm if needed.

charts/amazon-cloudwatch-observability/README.md
```diff
@@ -19,6 +19,129 @@ helm install --wait --create-namespace --namespace amazon-cloudwatch amazon-clou
...
+  finished requests are traced at all — aborts, timeouts and errors emit no span. So
+  Application Signals `Error` and `Fault` stay at zero regardless of what the engine is
+  doing. Upstream fix: [vllm#32162](https://github.com/vllm-project/vllm/pull/32162).
+- **No model name.** The V1 engine never sets `gen_ai.response.model`, so spans cannot be
+  joined to the `vllm:*` metrics by model. `service.name` carries `--served-model-name`.
+- **Traces are single-span.** vLLM honours an inbound W3C `traceparent`, but nothing in a
+  default Istio/Knative path propagates one, so there is no end-to-end trace and no
+  visibility into gateway queueing.
+
+### Enabling tracing on the engine
```
oberkem, line 107: it's okay if you want to submit another PR, but we can and should also get kserve traces https://knative.dev/docs/serving/observability/metrics/collecting-metrics/#enabling-metric-collection

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
```diff
@@ -375,6 +403,58 @@ processors:
...
+  filter/cw_k8s_ci_v0_kserve_controlplane_keep:
+    error_mode: ignore
+    metrics:
+      metric:
+        - 'not IsMatch(name, "^(controller_runtime_|workqueue_|rest_client_|leader_election_).*")'
```
oberkem, line 412: doesn't this drop a lot of other metrics? 
https://knative.dev/docs/serving/observability/metrics/serving-metrics/#workload-metrics
https://github.com/kserve/kserve/blob/master/python/kserve/kserve/metrics.py#L20

oberkem: Request-level telemetry is instead sourced from the Knative queue-proxy via knative.dataPlane, but that doesn't exist in RawDeployment mode afaik, so those users lose serving-layer request metrics entirely.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -120,6 +125,102 @@ receivers:
...
+            - role: pod
+              {{- if .Values.otelContainerInsights.solutions.vllm.namespace }}
+              namespaces:
+                names:
+                  - {{ .Values.otelContainerInsights.solutions.vllm.namespace }}
+              {{- end }}
+          relabel_configs:
+            # keep only KServe InferenceService pods (match the label VALUE, like
+            # the ebs-csi job keys on pod_label_app)
+            - source_labels: [__meta_kubernetes_pod_label_serving_kserve_io_inferenceservice]
```
oberkem, line 149: iiuc this job discovers targets by the serving.kserve.io/inferenceservice label and container name kserve-container, both of which only exist when the pod is reconciled from a KServe InferenceService. a plain vLLM Deployment (container vllm, no such label) gets zero targets and the vllm solution silently collects nothing. 
if the intention is to vLLM-on-KServe specifically note in values.yaml that solutions.vllm requires a KServe InferenceService. 
if the intention is vLLM generally, could we add a prometheus.io/scrape annotation fallback or any other fallback mechanism? 
I prefer we do VLLM generally but not sure if there would be specific things would be required for rayserve

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
Aakash-Dantre: Good catch. Fixed in 32c4649 on both jobs.

It was a real bug, not just inconsistency: without it every agent scraped every matching pod cluster-wide. One vLLM pod was emitting three copies of every `vllm:*` series under three `k8s.node.name` values — 3x cost, and `sum()` 3x high.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
Aakash-Dantre: Fixed in 32c4649 along with the vLLM job.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
Aakash-Dantre: Correct, and same as your Aug 26 comment — let me answer it.

A plain vLLM Deployment matches neither the label nor `kserve-container`, so it gets zero targets silently.

I'd rather do it your way. Proposal: scrape a pod if it's a KServe predictor **or** has `prometheus.io/scrape: "true"` + `prometheus.io/port`. Zero-config stays for KServe, plain Deployments get a standard opt-in.

RayServe also exposes `ray_serve_*` on the dashboard-agent port, so it needs its own solution rather than riding on this. OK to leave it out?

charts/amazon-cloudwatch-observability/values.yaml
Aakash-Dantre: Answered on the vLLM scrape job: https://github.com/aws-observability/helm-charts/pull/353#discussion_r4076476232 — short version, you're right, and I'd like to add a `prometheus.io/scrape` fallback so this isn't KServe-only.

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
Aakash-Dantre: Two things; the second is a real gap.

Neither link is emitted by the `kserve-controller-manager` — `metrics.py` is the KServe Python model server (predictor pod), and the Knative page is queue-proxy. I scraped the controller-manager directly: 158 families in, 21 kept, the rest `go_*`/`process_*`/`promhttp_*`.

Your follow-up is the real issue though: RawDeployment has no queue-proxy, so those users get no request telemetry from this PR. For vLLM the `http_*` series cover it (already kept by `vllm_keep`); for other runtimes we'd need a predictor scrape. Here or follow-up?

charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
Aakash-Dantre: Already there — `filter/cw_k8s_ci_v0_vllm_keep` at `_otel-container-insights-config.tpl:470`, wired in at :1220.

charts/amazon-cloudwatch-observability/README.md
Aakash-Dantre: Agreed, but the chart can't set it — `OTEL_TRACES_SAMPLER` is an env var on the engine container. I'll put `traceidratio` / `0.1` in the README next to `--otlp-traces-endpoint`.

Did "the whole thing" mean `traces.enabled: false`? That's off by default because without Transaction Search the endpoint 400s and the agent drops every span silently.

charts/amazon-cloudwatch-observability/README.md
Aakash-Dantre: I'll pick it up — blocked on conflicts and pre-commit, not review. It fixes span status and aborted requests (so Application Signals `Error`/`Fault` become real), but not `gen_ai.response.model`, which is https://github.com/vllm-project/vllm/issues/51186.

charts/amazon-cloudwatch-observability/README.md
Aakash-Dantre: Yes, as a follow-up. The valuable half is queue-proxy propagating W3C context into the engine — vLLM honours an inbound `traceparent` but nothing sends one today, so every trace is a single orphan span. I'll open an issue and link it.

PR #354, opened 2026-08-12
feat: selfTelemetry endpoint + watchReplicaset toggles

movence:
## What
- **`selfTelemetry.enabled`** (default `false`) + per-agent **`selfTelemetry.ports`**: injects `agent.self_telemetry` so each selected agent serves its own prometheus metrics on a loopback port.
- **`watchReplicaset`** toggle (default `true`), per pipeline:
  - `containerInsights.watchReplicaset` (ECI) → emits `watch_replicaset: false` into the generated agent config when off.
  - `otelContainerInsights.watchReplicaset` (OCI) → drops `k8s.deployment.name` from the k8sattributes extract (node + cluster-scraper) when off, stopping the ReplicaSet informer.

At their defaults, rendered manifests are unchanged.

## Testing
`helm lint` clean. `helm template` regression: with both toggles at defaults, rendered output is byte-identical to the base. Full matrix verified on a live EKS cluster — ECI-only / OCI-only / ECI+OCI × watchReplicaset{on,off} × selfTelemetry{on,off}: agent config correctness, pod health, loopback endpoint serving/closed as configured, and Container Insights metrics in CloudWatch.

## Merge order (3-repo stack) — this PR is #3
1. contrib — https://github.com/amazon-contributing/opentelemetry-collector-contrib/pull/623
2. agent — https://github.com/aws/amazon-cloudwatch-agent/pull/2246
3. chart (this PR) — https://github.com/aws-observability/helm-charts/pull/354

These toggles require an agent build that includes the agent PR, so this chart PR merges last.

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
+++ b/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
@@ -829,12 +829,14 @@ exporters:
 {{- end }}
 
 service:
-  {{- if not .Values.selfTelemetry.enabled }}
+  {{- $selfTelemetry := .Values.selfTelemetry | default dict }}
+  {{- $selfTelemetryPort := include "cloudwatch-agent.self-telemetry-port" (dict "agentName" .Values.otelContainerInsights.targetAgent "context" .) }}
+  {{- if not (and $selfTelemetry.enabled $selfTelemetryPort) }}
   # Pin internal self-telemetry off. Without this the collector falls back to its
   # built-in default (metrics level Normal + a Prometheus reader on localhost:8888),
   # which collides with the co-scheduled cluster-scraper on the node's :8888.
-  # When selfTelemetry is enabled it owns service.telemetry (with a non-8888 port), so this
-  # pin is omitted to avoid a merge race over the same key.
+  # Only omitted when THIS agent has a resolved self-telemetry port (so it actually owns
+  # service.telemetry on a non-8888 port); the global flag alone is not enough.
   telemetry:
     metrics:
       level: none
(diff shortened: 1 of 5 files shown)


charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```diff
@@ -827,12 +829,18 @@ exporters:
 {{- end }}
 
 service:
+  {{- $selfTelemetry := .Values.selfTelemetry | default dict }}
+  {{- $selfTelemetryPort := include "cloudwatch-agent.self-telemetry-port" (dict "agentName" .Values.otelContainerInsights.targetAgent "context" .) }}
+  {{- if not (and $selfTelemetry.enabled $selfTelemetryPort) }}
```
bhavya76: validate-flags (which includes the watchReplicaset and selfTelemetry.enabled type checks) is only called when the config has logs.metrics_collected.application_signals or logs.metrics_collected.kubernetes. What happens if we set selfTelemetry.enabled: "yes", Is it intentional to not error on that?

charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -1159,6 +1163,11 @@ applicationSignals:
 ##
 otelContainerInsights:
   enabled: false
+  ## Watch ReplicaSets to attribute pods to their Deployment via the k8sattributes
+  ## processor (adds k8s.deployment.name to the extract). Set to false to drop
+  ## k8s.deployment.name, which stops the k8sattributes ReplicaSet informer on every
```
bhavya76: Per the k8sattributes source (client.go), the ReplicaSet informer is gated on DeploymentUID, deployment label/annotation extraction, or ReplicaSet label/annotation extraction, not on k8s.deployment.name being in the metadata extract list. The toggle might only drop the attribute from the output, not affecting the informer or Role-Based Access Control

PR #358, opened 2026-08-26
Add GKE as a supported k8sMode

jefchien:
### Description of changes

Adds GKE (Google Kubernetes Engine) as a supported k8sMode in a similar way to the existing AKS support. GKE uses the same projected service-account token model as AKS.

Updates the OTel Container Insights to have GKE branches which set:
- `resourcedetection`: attaches `gcp` resource attributes
- `set_cloud_resource_id`: in the metrics and logs for the node and cluster-scraper to set the GKE cluster's Full Resource Name. See https://github.com/aws/amazon-cloudwatch-agent/pull/2258 for details (same setup in both).
- `deployment.environment.name`: evaluates to `{{cloud.platform}}:{{k8s.cluster.name}}/{{k8s.namespace.name}}` just like AKS
- API server endpoints setup matches AKS with `kubernetes.default.svc` workaround.

Includes a new `gcp-critical-pods` ResourceQuota just for GKE. GKE only admits pods that use the reserved `system-node-critical`/`system-cluster-critical` PriorityClasses in a namespace that has a ResourceQuota matching those classes.

https://developers.deepgram.com/docs/gcp-k8s#configure-namespace-resource-quotas-for-gke
> When deploying workloads in a non-default namespace (such as `dg-self-hosted`), GKE does not automatically provision quotas for the `system-node-critical` and `system-cluster-critical` priority classes in that namespace.

> Create a ResourceQuota in your Deepgram namespace to enable these priorities

> [!NOTE]
> Google's documentation unfortunately doesn't explicitly call out the behavior just that:
> https://docs.cloud.google.com/kubernetes-engine/quotas#resource_quotas
>> For clusters with under 100 nodes, GKE applies [Kubernetes resource quota](https://kubernetes.io/docs/concepts/policy/resource-quotas/) to every namespace
>
> Kubernetes documentation discusses the mechanisms of limiting priority class consumption by default, but nothing beyond that
> https://kubernetes.io/docs/concepts/policy/resource-quotas/#limiting-priorityclass-consumption-by-default
>> With this mechanism, operators are able to restrict usage of certain high priority classes to a limited number of namespaces and not every namespace will be able to consume these priority classes by default.

GKE configures and ships a `gcp-critical-pods` quota in its own managed namespaces.
```
NAMESPACE                           NAME                                      REQUEST       LIMIT   AGE
gke-managed-cim                     gcp-critical-pods                         pods: 1/1G            2d4h
gke-managed-networking-dra-driver   gke-managed-networking-dra-driver-quota   pods: 0/1G            2d4h
gke-managed-system                  gcp-critical-pods                         pods: 0/1G            2d4h
kube-system                         gcp-critical-pods                         pods: 22/1G           2d4h
```

The node agent, cluster-scraper agent, and fluent-bit all run as `system-node-critical` by default, so without the quota, the DaemonSets/Deployments fail to create the pods with `Error creating: insufficient quota to match these scopes: [{PriorityClass In [system-node-critical system-cluster-critical]}]`.

### Testing
With registered IAM OIDC Identity Provider and IAM role with trust policy set to:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "Federated": "arn:aws:iam::<ACCOUNT_ID>:oidc-provider/container.googleapis.com/v1/projects/<PROJECT_ID>/locations/us-east1-b/clusters/cwagent-otel-gke"
            },
            "Action": "sts:AssumeRoleWithWebIdentity",
            "Condition": {
                "StringEquals": {
                    "container.googleapis.com/v1/projects/<PROJECT_ID>/locations/us-east1-b/clusters/cwagent-otel-gke:sub": "system:serviceaccount:amazon-cloudwatch:cloudwatch-agent",
                    "container.googleapis.com/v1/projects/<PROJECT_ID>/locations/us-east1-b/clusters/cwagent-otel-gke:aud": "sts.amazonaws.com"
                }
            }
        }
    ]
}
```

Deployed to a GKE cluster and verified that the Helm chart works E2E.
```
helm upgrade --install amazon-cloudwatch-observability aws-observability/amazon-cloudwatch-observability \
  --namespace amazon-cloudwatch --create-namespace \
  --set k8sMode=GKE \
  --set roleArn=arn:aws:iam::<ACCOUNT_ID>:role/cwagent-gke-federation-test \
  --set region=us-east-1 \
  --set clusterName=cwagent-otel-gke \
  --set containerInsights.enabled=false \
  --set containerLogs.enabled=false \
  --set otelContainerInsights.enabled=true \
  --set otelContainerInsights.logs.enabled=true \
  --set-string 'agents[0].name=cloudwatch-agent' \
  --set-string 'agents[0].config=default:otel' \
  --set-string 'agents[1].name=cloudwatch-agent-cluster-scraper' \
  --set-string 'agents[1].mode=deployment' \
  --set-string 'agents[1].config=default' \
  --set agent.image.repositoryDomainMap.public=<ARTIFACT_REGISTRY_REPO> \
  --set agent.image.tag=<TAG>
```

> [!NOTE]
> Container insights application logs work as expected, but host logs are not collected on GKE. The host-log pipeline tails `/var/log/messages`, `/var/log/dmesg`, and `/var/log/secure`, but on the cluster's Container-Optimized OS nodes `/var/log/messages` and `/var/log/secure` don't exist and `/var/log/dmesg` is a directory (COS keeps system logs in the systemd journal, `/var/log/journal`). So the filelog receiver matches no files and the `/aws/otel/containerinsights/<cluster>/host` log group is never created. Application logs (from `/var/log/containers`) and metrics are unaffected.

image

---

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/gcp-critical-pods-resourcequota.yaml
+++ b/charts/amazon-cloudwatch-observability/templates/gcp-critical-pods-resourcequota.yaml
@@ -1,18 +1,18 @@
 {{- if eq .Values.k8sMode "GKE" }}
 {{- /*
-GKE only admits pods that use the reserved system-node-critical / system-cluster-critical PriorityClasses in a
-namespace that has a ResourceQuota matching those classes. The agent runs at system-node-critical, so without
-this quota its pods are rejected at admission on GKE. GKE-only. EKS and AKS do not gate priority classes this way.
+On GKE, a pod that uses the system-node-critical or system-cluster-critical PriorityClass is only admitted in a
+namespace that has a ResourceQuota with a scopeSelector that matches that class. The agent and fluent-bit run
+at system-node-critical by default.
 Ref: https://kubernetes.io/docs/concepts/policy/resource-quotas/#limit-priority-class-consumption-by-default
 */ -}}
 apiVersion: v1
 kind: ResourceQuota
 metadata:
+  labels:
+    {{- include "amazon-cloudwatch-observability.labels" . | nindent 4}}
   name: gcp-critical-pods
   namespace: {{ .Release.Namespace }}
 spec:
-  hard:
-    pods: "1G"
   scopeSelector:
     matchExpressions:
       - operator: In
--- a/charts/amazon-cloudwatch-observability/templates/gcp-critical-pods-resourcequota.yaml
+++ b/charts/amazon-cloudwatch-observability/templates/gcp-critical-pods-resourcequota.yaml
@@ -1,0 +1,23 @@
+{{- if eq .Values.k8sMode "GKE" }}
+{{- /*
+GKE only admits pods that use the reserved system-node-critical / system-cluster-critical PriorityClasses in a
+namespace that has a ResourceQuota matching those classes. The agent runs at system-node-critical, so without
+this quota its pods are rejected at admission on GKE. GKE-only. EKS and AKS do not gate priority classes this way.
+Ref: https://kubernetes.io/docs/concepts/policy/resource-quotas/#limit-priority-class-consumption-by-default
+*/ -}}
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+  name: gcp-critical-pods
+  namespace: {{ .Release.Namespace }}
+spec:
+  hard:
+    pods: "1G"
+  scopeSelector:
+    matchExpressions:
+      - operator: In
+        scopeName: PriorityClass
+        values:
+          - system-node-critical
+          - system-cluster-critical
+{{- end }}
--- a/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
+++ b/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-cluster-scraper-config.tpl
@@ -17,9 +17,9 @@ receivers:
           tls_config:
             ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
             insecure_skip_verify: false
-            {{- if eq .Values.k8sMode "AKS" }}
-            # Endpoints SD dials the apiserver by IP. On AKS the managed control-plane endpoint IP is
-            # not in the serving cert's SANs (only DNS names + the ClusterIP are), so verify against a
+            {{- if or (eq .Values.k8sMode "AKS") (eq .Values.k8sMode "GKE") }}
+            # Endpoints SD dials the apiserver by IP. On AKS and GKE the managed control-plane endpoint IP
+            # is not in the serving cert's SANs (only DNS names + the ClusterIP are), so verify against a
             # DNS SAN instead. On EKS the endpoint IPs are in the SANs, so this is not needed there.
             server_name: kubernetes.default.svc
             {{- end }}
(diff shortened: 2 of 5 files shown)


EduVencovsky (dismissed):

charts/amazon-cloudwatch-observability/templates/gcp-critical-pods-resourcequota.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if eq .Values.k8sMode "GKE" }}
+{{- /*
+GKE only admits pods that use the reserved system-node-critical / system-cluster-critical PriorityClasses in a
+namespace that has a ResourceQuota matching those classes. The agent runs at system-node-critical, so without
+this quota its pods are rejected at admission on GKE. GKE-only. EKS and AKS do not gate priority classes this way.
+Ref: https://kubernetes.io/docs/concepts/policy/resource-quotas/#limit-priority-class-consumption-by-default
+*/ -}}
+apiVersion: v1
+kind: ResourceQuota
+metadata:
```
movence, line 10: shoudl we add the default labels like other resources:
```
labels:
    {{- include "amazon-cloudwatch-observability.labels" . | nindent 4}}
```

charts/amazon-cloudwatch-observability/templates/gcp-critical-pods-resourcequota.yaml
```diff
@@ -0,0 +1,23 @@
...
+  name: gcp-critical-pods
+  namespace: {{ .Release.Namespace }}
+spec:
+  hard:
```
movence: is this required?

charts/amazon-cloudwatch-observability/templates/gcp-critical-pods-resourcequota.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if eq .Values.k8sMode "GKE" }}
+{{- /*
+GKE only admits pods that use the reserved system-node-critical / system-cluster-critical PriorityClasses in a
```
movence: does this mean that other pods like dcgm/neuron won't be scheduled unless they also get system-* PC?

charts/amazon-cloudwatch-observability/templates/gcp-critical-pods-resourcequota.yaml
jefchien: Makes sense. Adding.

charts/amazon-cloudwatch-observability/templates/gcp-critical-pods-resourcequota.yaml
jefchien: Doesn't seem like it. I included it to match the definitions of the other `gcp-critical-pods` resource quotas created for GKE managed namespaces, but removing it still allowed the pods to come up and it fits the example given in the k8s documentation.

https://kubernetes.io/docs/concepts/policy/resource-quotas/#limiting-priorityclass-consumption-by-default
```yaml
apiVersion: v1
kind: ResourceQuota
metadata:
  name: pods-cluster-services
spec:
  scopeSelector:
    matchExpressions:
      - operator : In
        scopeName: PriorityClass
        values: ["cluster-services"]
```

charts/amazon-cloudwatch-observability/templates/gcp-critical-pods-resourcequota.yaml
jefchien: No, the admission gating is only for pods that have `system-node-critical` or `system-cluster-critical` priority classes. DCGM and Neuron have neither set, so will be admitted and scheduled normally like every other pod.

jefchien: Tried to clarify the comment a bit.

movence (approved):

Paamicky (approved):

PR #360, opened 2026-08-31
Fix invalid quoting of daemonset maxUnavailable/maxSurge defaults

Paamicky:
### Description of changes:

We wrapped the DaemonSet `maxUnavailable/maxSurge` fields in `| toString |` quote. These are `IntOrString` fields, so the defaults rendered as "1"/"0" which is invalid, since a quoted value must be a percentage `([0-9]+%)`. This broke the default install for Fluent Bit and the CloudWatch Agent (daemonset mode).

### Changes
- Removed `| toString | quote` from these four lines so they render as valid bare integers, keeping all other string quoting intact:

- templates/linux/fluent-bit-daemonset.yaml
- templates/linux/cloudwatch-agent-custom-resource.yaml

### DIFF
--- a/charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
+++ b/charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
@@ -81,8 +81,8 @@ spec:
     type: {{ $agent.updateStrategy.type | quote }}
     {{- if eq $agent.updateStrategy.type "RollingUpdate" }}
     rollingUpdate:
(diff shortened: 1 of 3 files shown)


jefchien (dismissed):
nit: PR title is too generic. Please reword it so it's clear what this is fixing.

charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-custom-resource.yaml
```diff
@@ -81,8 +81,8 @@ spec:
     type: {{ $agent.updateStrategy.type | quote }}
```
jefchien, line 81: Looks like we'll need to modify https://github.com/aws-observability/helm-charts/blob/main/charts/amazon-cloudwatch-observability/tests/yaml_document_injection.sh as well.

PR #370, opened 2026-09-16
feat(dcgm): add g7e instance family to DCGM exporter node affinity

olowosulu:
## Summary

Adds the `g7e` EC2 instance family to the DCGM exporter's node-affinity instance-type allow-list (`dcgmExporter.affinity` in `values.yaml`), in both the plain and `ml.*` SageMaker blocks.

## Why

The allow-list uses `requiredDuringSchedulingIgnoredDuringExecution` with `operator: In` on `node.kubernetes.io/instance-type`. A node whose type is not in the list satisfies no `nodeSelectorTerm`, so the DaemonSet controller computes `desiredNumberScheduled = 0`, no exporter pod runs, and no GPU metrics reach CloudWatch Container Insights. This is what users on `g7e` nodes see today.

`g7e` (NVIDIA RTX PRO 6000 Blackwell Server Edition, GA 2026-01-20) is listed by AWS as a supported instance type for the EKS-optimized accelerated AMIs: https://docs.aws.amazon.com/eks/latest/userguide/ml-eks-optimized-ami.html

Same class of change as #258 (p6 families) and #109 (g6e).

## What is added

Exactly six sizes exist, confirmed via `ec2:DescribeInstanceTypes` in us-east-1, us-east-2 and us-west-2:

```
g7e.2xlarge  g7e.4xlarge  g7e.8xlarge  g7e.12xlarge  g7e.24xlarge  g7e.48xlarge
```

plus the matching `ml.g7e.*` entries. `g7e.xlarge` and `g7e.16xlarge` do **not** exist and are intentionally omitted. Entries are placed in family order between `g6e` and `gr6`, matching the existing layout.

## Deliberately not included

- `g6f` / `gr6f`: fractional-GPU instances (`GpuInfo.Gpus[0].Count = 0`, a fraction of one L4). NVIDIA documents DCGM support for "virtual machines with full GPU passthrough" (https://docs.nvidia.com/datacenter/dcgm/latest/installation.html), so DCGM behaviour on these is unconfirmed. Scheduling an exporter that may crashloop is worse than the current clean `DESIRED: 0`. Should be validated on a node before adding.
- `p6e-gb300`: could not verify the size suffix from a standard account, and #259 shows the cost of guessing (`u-p6e-gb200x*` had to be removed). Deferred until an authoritative size list is available.

## Validation

- `helm lint` passes (0 failed).
- `helm template` on the `DcgmExporter` CR: the only diff versus `main` is the 12 added lines; `In` list 141 → 153 values, 0 duplicates; the `eks.amazonaws.com/compute-type NotIn [fargate]` expression is unchanged.

## Note

A separate issue affects the current default DCGM image tag (`4.6.0-4.8.3-distroless`) on some nodes, where the exporter exits with `exit status 1` at startup — upstream: https://github.com/NVIDIA/dcgm-exporter/pull/724. This PR does not address that and the two should be considered together before a release, since scheduling the exporter onto g7e without the image issue resolved could surface it there.


charts/amazon-cloudwatch-observability/values.yaml
```diff
@@ -971,6 +977,12 @@ dcgmExporter:
                   - ml.g6e.24xlarge
                   - ml.g6e.48xlarge
                   - ml.g6e.xlarge
+                  - ml.g7e.2xlarge
```
movence: are these supported types with SageMaker? don't see them here https://docs.aws.amazon.com/sagemaker/latest/dg/notebooks-available-instance-types.html

movence (approved):

bhavya76 (approved):

